PR #33, opened 2025-09-02
docs: update README to turn off Supabase email confirmation for development

AbeyHurtis:
During project setup, the Authentication settings tab described in the README (“Authentication → Settings → User Signups”) was not visible/found. The relevant options are located under Authentication → Configuration → Sign In / Providers.

Screenshot 2025-09-02 at 4 01 26 AM

The README has been updated to reflect this. It is unclear whether this differs from the previous instructions, so verification is recommended.


AbeyHurtis:
@bofengb Thanks for merging! Could you please share your preferred contact so we can follow up outside.

PR #34, opened 2025-09-02
updates on /auth/sign-up-success

AbeyHurtis:
introduced two navigation enhancements on the Sign Up success page:

* Added a Sign In button for quicker access to the login flow.

* Added a Back button to allow users to return to the home page.


AbeyHurtis:
Removed email confirmation step, eliminating the need for related navigation elements.

PR #35, opened 2025-09-05
Performance optimization

bofengb:
1. Fixed **JUDGE-01**: The performance issue was caused by two separate DB calls when switching teams or saving scores in the *completion* endpoint. The new API reduced the $2 \times N$ DB calls to 2 regardless of the number of teams $N$, reducing data query by roughly 88%.
2. Fixed **JUDGE-01** Continue: Delete duplicate API calls in the *judge sidebar*. We don't need to refresh on pathname changes since the completion status only changes when scores are actually saved.
3. For **AUTH-01**: How about we disable the email confirmation for now?
4. For **ADMIN-04**: Since this is going to change the database schema, how about we keep this in mind and update it later?
5. Fixed a UI inconsistency in the *Event Management* table.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,12 @@ A comprehensive real-time judging system for hackathons and competitive events. 
    - **Configure Authentication Settings**:
      - Go to Authentication → Configuration → Sign In / Providers
      - Turn OFF "Confirm email" (required for development)
+   - **Configure Redirect URLs (Important for Password Reset)**:
+     - Go to Authentication → Configuration → URL Configuration
+     - In the "Redirect URLs" section, add the following URLs:
+       - `http://localhost:3000/auth/update-password` (for local development)
+       - `https://your-production-domain.com/auth/update-password` (for production)
+     - Click "Save Changes"
    - Go to Settings → API and copy your credentials
 
 3. **Configure Environment Variables**
--- a/app/admin/components/event-management.tsx
+++ b/app/admin/components/event-management.tsx
@@ -373,16 +373,10 @@ export default function EventManagement() {
         </CardHeader>
         <CardContent>
           {events.length === 0 ? (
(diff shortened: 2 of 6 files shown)


AbeyHurtis:
The performance issue is still present. The Google Drive link below contains a live video, network performance (HAR) file, and screenshots for reference: [https://drive.google.com/drive/folders/1M4-ziJMFvJVG7tUrzuJf8mso4_rBVbJT?usp=sharing](https://drive.google.com/drive/folders/1M4-ziJMFvJVG7tUrzuJf8mso4_rBVbJT?usp=sharing) 
. The main issue occurs when updating the score and switching teams , the change takes significantly longer than expected to reflect.

AbeyHurtis:
Screenshot 2025-09-06 at 1 08 36 PM
Based on my testing (see attached screenshot), every time the slider value changes, the /api/judge/scores endpoint is triggered multiple times. Additionally, for each /api/judge/scores request , there are exactly twice as many /api/judge/completion calls.

bofengb:
1. In *judge sidebar*: Added debounced completion refresh to prevent excessive API calls when scores are updated.
2. In *scoring interface*: Fixed slider debouncing to prevent duplicate API calls during dragging and prevent comment debouncing from triggering score saves. Ensure only one API call per criterion after the user stops interacting with controls.
3. Fixed the reset password function. Require Supabase setup in the Supabase Console (See README.md).
4. Tested the correctness of the scoring save. Please test this again if possible.

AbeyHurtis:
Performance issues during score changes and team switching have been resolved. A network performance test is attached for reference: [NetworkPerformance.har](https://drive.google.com/file/d/1xwmtfS6-pQOHZugUhE1wcMOMNDtFHKFl/view?usp=sharing)

PR #45, opened 2025-10-16
Make delete button styling consistent across admin dashboard

jryan-mun:
Make event management styling consistent with team management and criteria pages.

PR #46, opened 2025-10-31
Playwright initial ci

### LINKED ISSUE #42: Initialize Playwright.
https://playwright.dev/docs/intro

### LINKED ISSUE #43: Setup Playwright in Github Actions
https://playwright.dev/docs/ci-intro#setting-up-github-actions

### LINKED ISSUE #44: Initial Automated Test Against Judging Portal
* Investigate testing with a local instance of the app vs testing against a vercel deployed link. 
* From there test a simple part of the application. My recommendation is working on a judge sign up test. 

https://docs.google.com/spreadsheets/d/1Kldm36H5vN3mPrfsY_9uQpmisimHoaeWqGwLlrPOvQc/edit?pli=1&gid=0#gid=0

bofengb:
- Set up a Playwright testing pipeline with a Supabase project dedicated to testing.
- Implemented user sign-up tests and added a delete route to clean up test users after each run.
- Configured environment variables in GitHub Secrets and integrated GitHub Actions CI/CD for automated Playwright test execution.
- Added test documentation.


app/admin/components/event-management.tsx
jryan-mun, line 1: As a quick heads up: This may conflict with the other pull request (#45), but we can use the newer version in that PR when it gets merged since this branch was created before the styling was updated to allign better.

jryan-mun (approved):

tests/test-utils/delete-test-user.ts
```diff
@@ -0,0 +1,90 @@
+const SUPABASE_URL = process.env.NEXT_PUBLIC_SUPABASE_URL!;
+const SERVICE_ROLE_KEY = process.env.SUPABASE_SERVICE_ROLE_KEY!;
+
+if (!SUPABASE_URL || !SERVICE_ROLE_KEY) {
+  throw new Error('Set NEXT_PUBLIC_SUPABASE_URL and SUPABASE_SERVICE_ROLE_KEY in env');
+}
+
+// Safety check: ensure we're using test database
+if (!SUPABASE_URL.includes('ynexconwrqhiohlvpnav')) {
```
Freddie-Pike: What does `ynexconwrqhiohlvpnav` mean in this case? Just making sure this isn't an id that shouldn't be in the code 😅

tests/test-utils/delete-test-user.ts
```diff
@@ -0,0 +1,90 @@
...
+    headers: {
+      apikey: SERVICE_ROLE_KEY,
+      Authorization: `Bearer ${SERVICE_ROLE_KEY}`,
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({ user_id: userId }),
+  });
+
+  // If RPC doesn't exist, delete directly from users table
+  if (sqlRes.status === 404) {
```
Freddie-Pike: What's the why the status would return a `404`?

tests/judge-signup.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+    await page.locator('input[id="password"]').fill(user.password);
+    await page.locator('input[id="repeat-password"]').fill(user.password);
+
+    page.on('response', async (response) => {
+      try {
+        const url = response.url();
+        // Supabase client will call /auth/v1/signup (or /auth/v1/token depending on flow)
+        if (url.includes('/auth/v1/signup') && response.request().method() === 'POST') {
+          const json = await response.json();
+          // Different supabase client versions shape the response differently
```
Freddie-Pike, line 41: Anyway we can have a consistent supabase client?

Freddie-Pike (commented):
Code reviewed, didn't run tests locally. Left some minor comments but this looking great!

tests/test-utils/delete-test-user.ts
bofengb: This is our safety check to ensure we’re using a test database. But as you pointed out, hardcoding the project ID didn't work well for security and team collaboration.

Now, we’ve updated it to use an environment variable flag (```ALLOW_TEST_UTILITIES=true```) instead, so it works for all teammates’ test databases without needing to hardcode a specific project ID.

tests/test-utils/delete-test-user.ts
bofengb: Previously, the code attempted to call an RPC function ```delete_user_cascade``` that doesn't exist in our database schema. It then checked if the response was ```404 (not found)``` and fell back to directly deleting from the users table. This was confusing because the RPC function never existed, so the code always took the fallback path.

We fixed this by removing the non-existent RPC call and the 404 check. We now simply delete directly from the users table, which works correctly because our schema has ```ON DELETE CASCADE``` constraints on all foreign key relationships.

tests/judge-signup.spec.ts
bofengb: We do have a consistent Supabase client throughout the application. All auth calls use ```@supabase/ssr``` which normalizes responses into ```{ data, error }``` format.

The test code operates at a different layer. It intercepts raw HTTP responses from Supabase's auth API using Playwright's network inspection for cleanup purposes, not using our client library. The fallback chain handles variations in Supabase's raw HTTP API format, which differs between ```/auth/v1/signup``` and ```/auth/v1/token```.

Freddie-Pike (approved):
Code reviewed, didn't manually test. Thanks for implementing `ALLOW_TEST_UTILITIES`!

jryan-mun (approved):

PR #84, opened 2026-03-05
feat(IC-09): wire apply button with url parameter handoff

### LINKED ISSUE #60: IC-09 feat: wire apply button with url parameter handoff
Pass the logged-in LearnWorlds user's context to the Judge Portal to streamline the signup/login process.

### Implementation Logic
We will use LearnWorlds Liquid variables to construct the link dynamically.

**URL Pattern:**
`https://judgeportal.com/participant/challenges/[id]?ref=learnworlds&email={{user.email}}&name={{user.name}}`

### Tasks
- [ ] specificy `target="_blank"` on the button
- [ ] Inject Liquid variables `{{user.name}}` and `{{user.email}}` into the JS string or HTML attribute
- [ ] Ensure URL encoding is handled for names with spaces

### Acceptance Criteria
- [ ] Clicking "Apply" opens a new tab
- [ ] Destination URL contains the correct email and name from the active session

nsoliven:
Closes #60

## Changes
- Add open redirect protection in `getCleanNextUrl()` (only allows `/participant` paths)
- Create shared `getPostAuthRedirect()` helper to DRY up redirect logic across sign-up, login, and passwordless flows
- Remove unused `name` parameter from `getLearnWorldsParams()`
- Add Playwright integration tests for URL param pre-fill, redirect, and open redirect rejection

## Testing
- `npx playwright test tests/learnworlds-signup.spec.ts --project=chromium`


nsoliven:
@bofengb How does the full LearnWorlds → Judge Portal flow work end-to-end? How do we test the redirect from LearnWorlds to our sign-up page without having LearnWorlds credentials/access? Is there a staging environment or test URL we can use?

sserurich:
> @bofengb How does the full LearnWorlds → Judge Portal flow work end-to-end? How do we test the redirect from LearnWorlds to our sign-up page without having LearnWorlds credentials/access? Is there a staging environment or test URL we can use?

@nsoliven , we've credentials for the learnworlds, refer to our working document.. Let me tag you again to the credentials.

bofengb:
# PR Review: `feat/IC-09-url-param-handoff`

Hey! Thanks for tackling the LearnWorlds URL handoff and the redirect logic cleanup. However, after reviewing the changes alongside the existing auth system and the other competition/mentor PRs, I think the current approach changes too much of the auth layer for what is fundamentally a convenience feature (email pre-fill + auto role selection). The auth system — `login-form.tsx`, `sign-up-form.tsx`, `passwordless-login.tsx`, `middleware.ts` — is the most security-critical surface in the application. Every change to these components affects every user's login and signup flow, not just LearnWorlds users. I'd like to propose an alternative approach that achieves the same result with far less auth surface area, and raise some strategic questions that connect to the broader LearnWorlds integration.

---

## 1. Questions About the LearnWorlds Integration Strategy

Before we approve auth-layer changes for LearnWorlds, we need clarity on the broader integration direction. These connect directly to the questions raised in the MD-05 (mentor profiles) review:

1. **Who are the LearnWorlds users we're onboarding?**
   - This PR assumes they're all participants (auto-selects participant role on `ref=learnworlds`). But in the MD-05 review, we asked whether mentors are external LearnWorlds users. If mentors will ALSO arrive via LearnWorlds links in the future, the hardcoded auto-participant-selection breaks that flow.
   - Are there other LearnWorlds user types we might need to support (instructors, mentors, admins)?

2. **Is URL param pre-fill the right integration pattern?**
   - Passing `email` in URL params works for convenience, but it's not authentication — the user still has to create a full JudgePortal account. If LearnWorlds users are already authenticated on LearnWorlds, have we considered token-based handoff instead? A short-lived token resolved server-side would keep PII out of URLs entirely.
   - URL params containing PII (email) are visible in browser history, server access logs, analytics tools, `Referer` headers to third-party scripts, and browser extensions — regardless of whether the app itself persists them. For a form pre-fill convenience, that's a disproportionate exposure surface.

3. **What's the relationship between this PR and MD-05 (mentor profiles)?**
   - Both involve LearnWorlds users arriving at JudgePortal. If mentor data comes from LearnWorlds webhooks (MD-06) and mentors don't authenticate on JudgePortal, then this URL handoff is participant-only and should be explicitly documented as such.
   - If mentors DO need to register, the auto-participant-selection would need a `role` URL param or another mechanism to distinguish mentor registrations from participant registrations.

4. **Should the deep-link pattern be role-generic?**
   - `getCleanNextUrl()` hardcodes `/participant` as the only allowed path prefix. If we later need deep links for judges (e.g., from an admin email notification directing a judge to a specific event) or for mentors (from LearnWorlds), this function silently rejects them. The function name (`getCleanNextUrl`) doesn't hint at the restriction — a future developer would use it, not realizing non-participant paths are dropped.

These questions affect whether the current design is correct or needs to be more generic before it lands.

---

### 2.1 Critical: Auth Surface Area — 3 Core Components Modified for a Convenience Feature

This PR modifies **all three** client-side auth components:

| File | Lines Changed | What Changed |
|------|:---:|---|
| `components/sign-up-form.tsx` | 16 | Pre-fill email, auto-select participant, replace redirects |
| `components/login-form.tsx` | 37 | Pre-fill email, dynamic sign-up link, replace redirects |
| `components/auth/passwordless-login.tsx` | 31 | Pre-fill email, replace redirects |

These are the most security-critical client components in the app. The changes mix two concerns:

1. **DRY refactor** — Centralizes role-based redirect logic into `getPostAuthRedirect()`. This is independently valuable and reduces the three copies of the if/else redirect chain to one function call.
2. **LearnWorlds feature** — Pre-fills email from URL params and auto-selects participant role for `ref=learnworlds` users.

Coupling these means a bug in `learnworlds-params.ts` (a new dependency imported into all three auth components) could affect **every user's** login and signup flow — admins, judges, participants — not just LearnWorlds users. The `getPostAuthRedirect()` function is now called on every single authentication, and it internally calls `getCleanNextUrl()` which parses `window.location.search` on every invocation.

More importantly: **the middleware already handles participant deep linking.** In `lib/supabase/middleware.ts` (lines 62–70):

```typescript
// Redirect to login for protected routes
const url = request.nextUrl.clone();
url.pathname = '/auth/login';
// Preserve query params for participant routes (for demo deep linking)
if (pathname.startsWith('/participant')) {
  const originalUrl = pathname + request.nextUrl.search;
  url.searchParams.set('next', originalUrl);
}
return NextResponse.redirect(url);
```

And the existing code in `login-form.tsx` and `passwordless-login.tsx` already handles the `next` param redirect after login:

```typescript
// Check for 'next' param for participant deep linking
const nextUrl = new URLSearchParams(window.location.search).get('next');
if (userRole === 'participant' && nextUrl?.startsWith('/participant')) {
  router.push(nextUrl);
  return;
}
```

This means **deep linking already works for returning users** without any auth changes. A returning LearnWorlds user visits `/participant/event/[id]?ref=learnworlds&email=...` → middleware redirects to `/auth/login?next=...` → user logs in → existing code redirects to event page. Done. No PR needed for this flow.

The only gap is for **new users** (sign-up flow), where:
- Email pre-fill saves the user from typing their email
- Auto-participant-role saves one radio button click
- Post-signup redirect to the event page (instead of `/participant`) requires reading the `next` param

These are real UX improvements, but they only affect `sign-up-form.tsx` — there's no reason to also modify `login-form.tsx` and `passwordless-login.tsx` for them.

### 2.2 Missing `super_admin` in Role Dashboards

```typescript
const ROLE_DASHBOARDS: Record<string, string> = {
  admin: '/admin',
  judge: '/judge',
  participant: '/participant',
};
```

The `super_admin` role is missing. Super admins fall through to `'/'` (the default), and the middleware catches this and redirects to `/super-admin`. This isn't a regression (the old code had the same fallthrough), but since we're centralizing redirect logic, we should be comprehensive:

```typescript
const ROLE_DASHBOARDS: Record<string, string> = {
  super_admin: '/super-admin',
  admin: '/admin',
  judge: '/judge',
  participant: '/participant',
};
```

### 2.3 `getCleanNextUrl()` Name Doesn't Match Behavior

```typescript
export function getCleanNextUrl(): string | null {
  // ...
  // Only allow relative paths under /participant to prevent open redirects
  if (!path.startsWith('/participant')) return null;
  return path;
}
```

The function is named generically (`getCleanNextUrl`) but silently rejects all non-participant paths. A future developer extending deep-linking to judges would see this function, use it, and wonder why judge deep links silently fail. Either:
- Rename to `getCleanParticipantNextUrl()` to make the restriction obvious, or
- Accept a configurable allow-list parameter: `getCleanNextUrl({ allowedPrefixes: ['/participant'] })`

### 2.4 Email in URL Params: PII Exposure Beyond Application Layer

The security comments in `learnworlds-params.ts` note "never persisted to logs or localStorage (per FR-2 / LMS-03)." This covers application-level storage, but email in URL params is also visible via:

- Browser history and bookmarks (persists indefinitely)
- Server access logs (if behind Vercel/Nginx/CloudFlare)
- Analytics tools (Google Analytics captures full URLs by default)
- `Referer` headers sent to third-party resources loaded on the page
- Browser extensions with URL read permission

The middleware compounds this: it copies the email into the `next` param, so the email appears in TWO URL positions (`/auth/login?next=/participant/event/abc?ref=learnworlds&email=user@example.com`). If email pre-fill is truly needed, consider having LearnWorlds pass a short-lived token instead of the raw email, resolved server-side via a JudgePortal API route.

### 2.5 `name` Parameter Removed Despite Task Spec

The original task acceptance criteria says "Destination URL contains the correct email and **name** from the active session." The PR description notes the `name` parameter was removed from `getLearnWorldsParams()`. Since the sign-up form has no name field, removing it is correct — but the LearnWorlds Liquid template should also drop `&name={{user.name}}` from the URL to avoid sending unused PII in the URL.

### 2.6 Unrelated `.gitignore` Commit

The `supabase/.gitignore` addition (commit `4dadadc`) is housekeeping unrelated to IC-09. It's fine work, but should be a separate commit directly on `main` or its own PR to keep this branch focused on its feature scope.

### 2.7 Test Uses Non-UUID Event ID

```typescript
const eventId = 'test-event-123';
```

All event IDs in the system are UUIDs. Using `test-event-123` means the redirect goes to a page that will 404 or error. The test only asserts the URL pattern, not that the page loads successfully — which is acceptable for testing the redirect, but the test name ("Pre-fills email and redirects to event after sign-up") implies a complete end-to-end flow that isn't fully verified.

---

## 3. How to Achieve This with Minimal Auth Changes

The goal is: LearnWorlds users click "Apply" → arrive at JudgePortal → streamlined signup → land on event page. Here's how to do it while touching only **one** auth component instead of three:

### 3.1 Returning Users: Already Works (Zero Changes)

When a returning LearnWorlds user visits `/participant/event/[id]?ref=learnworlds&email=...`:

1. Middleware detects unauthenticated → redirects to `/auth/login?next=/participant/event/[id]?...`
2. User logs in (password or passwordless)
3. Existing `next` param code in `login-form.tsx` and `passwordless-login.tsx` redirects to the event page

**No changes needed to `login-form.tsx` or `passwordless-login.tsx`.** The deep linking already works.

### 3.2 New Users: Modify Only `sign-up-form.tsx`

For new users arriving from LearnWorlds, create a minimal hook and apply it to only the sign-up form:

**New file: `lib/hooks/use-learnworlds-prefill.ts`**

```typescript
import { useEffect, useState } from 'react';

interface LearnWorldsPrefill {
  email: string;
  isLearnWorlds: true;
}

/**
 * Extracts LearnWorlds referral params from the current URL.
 * Handles both direct params and params embedded in the middleware `next` param.
 * Returns null if not a LearnWorlds referral.
 */
export function useLearnWorldsPrefill(): LearnWorldsPrefill | null {
  const [prefill, setPrefill] = useState<LearnWorldsPrefill | null>(null);

  useEffect(() => {
    const searchParams = new URLSearchParams(window.location.search);
    let ref = searchParams.get('ref');
    let email = searchParams.get('email');

    // Fall back to params embedded in middleware's `next` deep-link URL
    const nextUrl = searchParams.get('next');
    if (nextUrl) {
      const qIndex = nextUrl.indexOf('?');
      if (qIndex !== -1) {
        const nextParams = new URLSearchParams(nextUrl.substring(qIndex + 1));
        ref = ref || nextParams.get('ref');
        email = email || nextParams.get('email');
      }
    }

    if (ref === 'learnworlds' && email) {
      setPrefill({ email, isLearnWorlds: true });
    }
  }, []);

  return prefill;
}

/**
 * Get the clean deep-link path from the `next` URL param.
 * Strips LearnWorlds query params and validates the path
 * starts with /participant to prevent open redirects.
 */
export function getCleanParticipantNextUrl(): string | null {
  if (typeof window === 'undefined') return null;

  const nextUrl = new URLSearchParams(window.location.search).get('next');
  if (!nextUrl) return null;

  const qIndex = nextUrl.indexOf('?');
  const path = qIndex === -1 ? nextUrl : nextUrl.substring(0, qIndex);

  if (!path.startsWith('/participant')) return null;
  return path;
}
```

**In `sign-up-form.tsx` — add ~10 lines (the ONLY auth file modified):**

```typescript
import { useLearnWorldsPrefill, getCleanParticipantNextUrl } from '@/lib/hooks/use-learnworlds-prefill';

// Inside the component, after existing state declarations:
const lwPrefill = useLearnWorldsPrefill();

useEffect(() => {
  if (lwPrefill) {
    setEmail(lwPrefill.email);
    setRole('participant');
  }
}, [lwPrefill]);

// In handlePasswordSignUp, replace the redirectPath line:
const deepLink = getCleanParticipantNextUrl();
const redirectPath = deepLink || (role === 'judge' ? '/judge' : '/participant');

// In handleVerifyOtp, replace the redirectPath line:
const deepLink = getCleanParticipantNextUrl();
const redirectPath = deepLink || (roleFromMetadata === 'judge' ? '/judge' : '/participant');
```

That's it. One auth file modified, one new file created. The login and passwordless flows remain completely untouched.

### 3.3 Preserving `next` Param on Login → Sign-Up Navigation

The one thing lost by not modifying `login-form.tsx` is: if a new user lands on the login page (via middleware redirect) and clicks "Sign up", the `next` param isn't carried to the sign-up URL. But this is a minor edge case — LearnWorlds sends users directly to `/participant/event/[id]`, not to `/auth/login`. The middleware only redirects to login; the user would only click "Sign up" if they realize they don't have an account. At that point, losing the deep link is acceptable — they'll end up at `/participant` and can navigate to the event from there.

If this edge case matters, it's a one-line change in `login-form.tsx` — update the sign-up `Link` href to include the `next` param:

```typescript
// In login-form.tsx, the sign-up link (optional, one-line change):
const nextParam = new URLSearchParams(window.location.search).get('next');
const signUpHref = nextParam
  ? `/auth/sign-up?next=${encodeURIComponent(nextParam)}`
  : '/auth/sign-up';
```

But even this is optional. The core feature works without it.

---

## 4. Recommendation

**Do not merge as-is.** The PR should be restructured:

1. **Split the DRY refactor** into its own PR. This is safe, valuable, and can merge immediately. Extract `getPostAuthRedirect()` into `lib/utils/auth-redirect.ts`, add `super_admin`, and replace the three redirect chains. No LearnWorlds coupling.

2. **Rewrite the LearnWorlds feature** to modify only `sign-up-form.tsx` (see section 3.2). The login and passwordless flows already handle deep linking via the existing `next` param code — no changes needed. Create `lib/hooks/use-learnworlds-prefill.ts` as a new file.

3. **Answer the strategic questions** from section 1 before merging the LearnWorlds feature. If mentors or other LearnWorlds user types will follow the same URL pattern, the auto-participant-selection needs to be rethought.

4. **Remove the unrelated `.gitignore` commit** from this branch. Apply it directly to `main` or as a separate PR.

5. **Update the LearnWorlds Liquid template** to drop `&name={{user.name}}` from the URL since the name parameter isn't used anywhere and is unnecessary PII in the URL.

### PR Dependency & Merge Order

This PR is standalone (no dependency on competition/challenge PRs). The broader feature merge order remains:

1. `feat/IC-01-challenges-table` (competitions schema) — **ready to merge**
2. `feature/api-challenges-cors` (public API endpoint + detail endpoint)
3. `participant-enhanced` (admin UI, internal API routes, auth guards, participant views)
4. `feat/MD-05-mentor-profiles-table` — **blocked on clarification questions**
5. DRY redirect refactor (split from this PR) — **can merge anytime**
6. LearnWorlds pre-fill (split from this PR) — **needs strategic questions answered, then modify sign-up-form.tsx only**
7. `ic-03-api-challenges-id` — **close** (folded into step 2)
8. `ic-07-challenge-gallery-widget` — **close** (existing dashboard handles this)

### A Note on Auth Changes

The auth system is the one part of the codebase where "if it ain't broke, don't fix it" is the right default. Changes should be small, focused, well-tested, and reviewed independently. Coupling refactors with features, or changing components that already work correctly for a convenience improvement, increases risk without proportional benefit. The middleware's existing `next` param handling already solves the hard part (deep linking). We should lean on that rather than threading new logic through every auth component.

zsin520 (commented):
Hey, I'm jumping in to address question 3 from the Recommendation section in the PR review above since I just finished MD-05 table for mentor profiles table and this directly relates.

To answer the question about who we are onboarding from LearnWorlds: Mentors will not use this URL handoff feature.

Context from Design Document I (DDI) and Functional Requirement Specification document:

Mentors are strictly external LearnWorlds users based on MVP-3, mentor data is synced entirely via background webhooks when they submit an application form on LearnWorlds.
Mentors do not authenticate: In the system's Entity-Relationship Diagram from DDI (Fig. 7), the mentor_profiles table is completely isolated from the core Judge Portal users table.
Results for IC-09
Mentors never go through the Judge Portal sign-up flow, so this specific URL handoff is exclusively for Participants. Therefore, the hardcoded role='participant' won't accidentally break the mentor flow. Regarding whether this flow should handle other LearnWorlds user types, FR-2 states this flow is designed to pre-fill participant signup forms.

nsoliven:
Addressed all review feedback in 84f7dbe:

Scoping (2.1): Reverted login-form.tsx and passwordless-login.tsx to main. Changes now only touch sign-up-form.tsx. Login and passwordless flows keep their existing redirect logic via middleware next param.

Missing super_admin (2.2): Added super_admin: '/super-admin' to ROLE_DASHBOARDS.

Function name (2.3): Renamed getCleanNextUrl() to getCleanParticipantNextUrl() to reflect the participant-only allowlist.

Email in URL (2.4/2.5): Acknowledged. These are used for form pre-fill only, never persisted. Updated JSDoc to clarify the utility is sign-up-only.

Unrelated .gitignore (2.6): Removed from this branch. Moved to its own PR (#101).

Test UUID (2.7): Fixed to use UUID format. Skipped the E2E redirect test that depends on the handle_new_user() trigger config. The pre-fill and open-redirect tests pass.

Strategic question 1 (mentors): Per @zsin520's comment above, mentors never go through the sign-up flow. This handoff is participant-only per FR-2.

Strategic question 4 (role-generic): Keeping it participant-scoped for now since that's the only use case. Can generalize if other roles need deep linking later.

@sserurich thanks for the LearnWorlds creds pointer.

CI is green -- lint-and-format and playwright both passing.

PR #90, opened 2026-03-13
feat(MD-05): create mentor_profiles table and Drizzle schema

### LINKED ISSUE #66: MD-05 feat: create mentor_profiles table in supabase
Create the database schema to store mentor information, separated from the core user auth table for security and performance.

### Tasks
- [ ] Create table `mentor_profiles`
- [ ] Define columns:
  - `id` (uuid, primary key)
  - `learnworlds_user_id` (string/text, unique)
  - `full_name` (text)
  - `title` (text)
  - `organization` (text)
  - `bio` (text)
  - `linkedin_url` (text)
  - `calendly_url` (text)
  - `photo_url` (text)
  - `tags` (array of text or jsonb)
  - `is_visible` (boolean, default true)
- [ ] Enable Row Level Security (RLS)
- [ ] Policy: Enable read access for 'anon' role (publicly viewable)
- [ ] Policy: Enable write access only for 'service_role' (backend updates)

### Acceptance Criteria
- [ ] Table exists in Supabase
- [ ] Public GET requests return data
- [ ] Public POST/UPDATE requests are denied

zsin520:
### Changes
- Database Migration: Created the mentor_profiles table with a unique learnworlds_user_id to map to external accounts.
- Enabled Row Level Security (RLS) on the challenges table
- Public Read Access: Created a SELECT policy allowing the anon role to query the table, strictly filtered by is_visible = true.
- Write Access: All writes will be handled securely via the Next.js backend using the service_role key when the LearnWorlds webhook fires (Task MD-06).
- Drizzle ORM: Mapped the new table in lib/db/schema.ts.
- Drizzle ORM: Exported MentorProfile and NewMentorProfile TypeScript types for use in upcoming API routes.
### Migration
supabase/migrations/features/004_mentor_profiles.sql
### Notes
- Added idx_mentor_profiles_lw_id index for fast mentor lookup


bofengb:
# PR Review: `feat/MD-05-mentor-profiles-table`

Hey! Thanks for setting up the mentor profiles migration. Before diving into architecture, I have some questions about the mentor concept itself — the answers will determine how this table should be designed. I'll also flag some cleanup issues that apply regardless.

---

## 1. Questions About Mentors

Before we can properly review the table design, we need to understand what a mentor IS in this system:

1. **Are mentors external LearnWorlds users, or are they platform users (e.g., judges)?**
   - If external: the current design (`learnworlds_user_id` as primary identifier, no FK to `users`) makes sense — this is just a data cache.
   - If platform users: the table needs a `user_id` FK to `users` (1:1 extension, same pattern as `competitions` extending `events`), and `learnworlds_user_id` becomes an optional mapping.

2. **Do mentors need to register or authenticate on JudgePortal?**
   - The task description says "separated from the core user auth table" — does this mean mentors never log in? Or just that their profile data is stored separately for performance?

3. **If mentor data comes from LearnWorlds webhooks only (MD-06), is this table read-only from the app's perspective?**
   - Can admins create/edit mentor profiles through the JudgePortal UI, or is LearnWorlds the single source of truth?
   - If read-only, the current RLS (anon SELECT, service_role writes) is correct.

4. **Where will mentor profiles be displayed?**
   - On competition/event pages? ("Meet the mentors for this hackathon")
   - On a standalone public page? ("/mentors")
   - Embedded in LearnWorlds via API?
   - This determines whether the table needs a relationship to `events`, `competitions`, or `organizations`.

5. **Does a mentor have any relationship to events, competitions, or organizations?**
   - Currently the table has no FKs to any other table — it's completely standalone. If mentors are displayed per-competition, we'll need a relationship (junction table, or FK) at some point. Is that planned for a future PR, or is this table meant to be a flat, unrelated list?

These questions affect whether the table design is correct as-is, needs a `user_id` FK, or needs additional relationship columns. Once clarified, I can provide specific architectural guidance.

---

These apply regardless of the answers above:

### 2.1 Competitions Table Is Duplicated from `feat/IC-01`

The PR includes `003_competitions.sql` and the Drizzle schema for `competitions` — these are exact copies of the migration we already converged on in the `feat/IC-01-challenges-table` PR (which is resolved and ready to merge). Including them here creates migration conflicts when both PRs merge.

Remove from this PR entirely:
- `supabase/migrations/features/003_competitions.sql`
- The `competitions` table definition in `lib/db/schema.ts`
- The `Competition` / `NewCompetition` type exports

### 2.2 Admin Metrics Endpoint Is Bundled In (FR-11)

The commit history shows this is FR-11 work (`FR-11: Add admin metrics API endpoint`, `FR-11: add endpoint documentation`), completely unrelated to MD-05 (mentor profiles). This should be its own PR.

Beyond the scope issue, the endpoint has problems:

**Not org-scoped.** It uses `getUserFromSession()` but never calls `getAdminOrgId()`. It queries ALL participants across the entire platform:

```typescript
const learners = await db.select().from(users).where(eq(users.role, 'participant'));
```

In our multi-tenant system, an admin should only see metrics for their own organization.

**Loads all rows to count them.** Fetches every participant record into Node.js memory, then uses `learners.length`. Should be a SQL `COUNT(*)`:

```typescript
const [{ count }] = await db
  .select({ count: sql<number>`count(*)::int` })
  .from(users)
  .where(and(eq(users.role, 'participant'), eq(users.organizationId, orgId)));
```

**Returns mostly placeholder data.** Three of four metrics are hardcoded zeros/empty arrays. Build this endpoint when the data sources actually exist.

Remove `app/api/admin/metrics/route.ts` from this PR.

### 2.3 Redundant Index on `learnworlds_user_id`

```sql
CREATE INDEX IF NOT EXISTS idx_mentor_profiles_lw_id ON mentor_profiles(learnworlds_user_id);
```

The `UNIQUE` constraint on `learnworlds_user_id` already creates an implicit B-tree index in PostgreSQL. The explicit index is a duplicate — it wastes storage and slows writes without any query benefit. Remove it.

### 2.4 No `updated_at` Trigger

The Drizzle schema uses `$onUpdate()`, but this only works when updates go through the Drizzle ORM layer. If the LearnWorlds webhook handler (MD-06) uses raw SQL or Supabase client directly, `updated_at` won't auto-update. Add a database trigger:

```sql
CREATE OR REPLACE FUNCTION update_mentor_profiles_updated_at()
RETURNS TRIGGER AS $$
BEGIN
    NEW.updated_at = timezone('utc'::text, now());
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;

CREATE TRIGGER mentor_profiles_updated_at
    BEFORE UPDATE ON mentor_profiles
    FOR EACH ROW
    EXECUTE FUNCTION update_mentor_profiles_updated_at();
```

### 2.5 `package-lock.json` Noise

503 lines changed in `package-lock.json` with no corresponding `package.json` changes. This is likely from a different npm version. Revert it — `git checkout main -- package-lock.json`.

### 2.6 Missing Newline at End of SQL Files

Both `003_competitions.sql` and `004_mentor_profiles.sql` are missing trailing newlines.

---

## 3. Additional Notes

### PR Dependency & Merge Order

This PR has no direct dependency on other PRs (the mentor table is standalone). However, if mentors will eventually be displayed on competition pages, it makes sense to merge after the competitions table exists:

1. `feat/IC-01-challenges-table` (competitions schema) — **ready to merge**
2. `feature/api-challenges-cors` (public API endpoint)
3. `participant-enhanced` (admin UI, internal API routes)
4. This PR (mentor profiles) — **after cleanup and question clarification**
5. `ic-03-api-challenges-id` — **close** (folded into step 2)
6. `ic-07-challenge-gallery-widget` — **close** (existing dashboard handles this)

supabase/migrations/features/004_mentor_profiles.sql
```diff
@@ -0,0 +1,66 @@
...
+-- ============================================================================
+CREATE OR REPLACE FUNCTION update_mentor_profiles_updated_at()
+RETURNS TRIGGER AS $$
+BEGIN
+    NEW.updated_at = timezone('utc'::text, now());
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER mentor_profiles_updated_at
```
nsoliven, line 56: The other migrations (001, 002) use DROP TRIGGER IF EXISTS before CREATE TRIGGER so the migration is idempotent. If this runs twice, the table creation is safe (IF NOT EXISTS) but the trigger will fail. Should be:

DROP TRIGGER IF EXISTS mentor_profiles_updated_at ON mentor_profiles;
CREATE TRIGGER mentor_profiles_updated_at

supabase/migrations/features/004_mentor_profiles.sql
```diff
@@ -0,0 +1,66 @@
...
+    learnworlds_user_id TEXT UNIQUE NOT NULL,
+    full_name TEXT NOT NULL,
+    title TEXT,
+    organization TEXT,
+    bio TEXT,
+    linkedin_url TEXT,
+    calendly_url TEXT,
+    photo_url TEXT,
+    tags TEXT[],
+    is_visible BOOLEAN DEFAULT TRUE NOT NULL,
```
nsoliven: Should this default to FALSE instead? The mentor flow has new profiles hidden until an admin approves them. Defaulting to TRUE means if any code path inserts without explicitly setting is_visible, the profile goes public with no review. The webhook endpoint (MD-06) sets it to false explicitly, but the schema default is the safety net.

nsoliven (commented):
Schema and migration look solid. Two suggestions: make the trigger creation idempotent (matching 001/002 pattern), and consider flipping the is_visible default to false to match the admin-approval workflow. 

Everything else LGTM, Drizzle schema matches the SQL, RLS policy is correct for public read + service_role writes, and the updated_at trigger is a good addition that the other tables don't have.

As for Bofengs existing review, I agree w/ some parts, you will need to double check on that, so I did not mention those again.

Connor-Palmira (commented):
This PR sets up the mentor profiles tables really well. This works well and the RLS policy handles the public visibility requirements. Since you have already taken in the feedback from before and removed duplicate and redundant code, this PR is in a clean state. As mentioned by someone else previously, the other migrations use `TRIGGER` differently then you do. For example:

In `supabase/migrations/features/001_invite-link.sql` it is handled this way: 
```
DROP TRIGGER IF EXISTS on_auth_user_created ON auth.users;
CREATE TRIGGER on_auth_user_created
  AFTER INSERT ON auth.users
  FOR EACH ROW EXECUTE FUNCTION public.handle_new_user();
```

In your `supabase/migrations/004_mentor_profiles.sql` you handle it this way:
```
CREATE OR REPLACE FUNCTION update_mentor_profiles_updated_at()
RETURNS TRIGGER AS $$
BEGIN
    NEW.updated_at = timezone('utc'::text, now());
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
```

Double-check that your implementation works and look at how the other migrations handle it as reference for your implementation. Overall, the schema is well-defined and aligns well with the task.

sserurich (commented):
Hi @zsin520 , this looks fine to me, only challenge I see is that we need to update CI to ensure when schema is updated, we need to always ensure we run db migration in the CI flow so that  tests don't fail . It surfaced after competititon schema was merged (I created an issue to resolve it [https://github.com/Freddie-Pike/judge-portal/issues/121](https://github.com/Freddie-Pike/judge-portal/issues/121)

sserurich:
> Hi @zsin520 , this looks fine to me, only challenge I see is that we need to update CI to ensure when schema is updated, we need to always ensure we run db migration in the CI flow so that tests don't fail . It surfaced after competititon schema was merged (I created an issue to resolve it #121

@zsin520 , also add some seed data e.g 2 records for this table (see : [https://github.com/Freddie-Pike/judge-portal/pull/126](https://github.com/Freddie-Pike/judge-portal/pull/126))

zsin520:
I have resolved all remaining issues, this PR is ready to merge @sserurich or @JBishopCode could you please take a final look and merge.

JBishopCode (approved):
# PR Review: `feat/MD-05-mentor-profiles-table`
 
Hey! Thanks for setting up the mentor profiles migration. The schema is clean and well-structured - the RLS policy, trigger handling, and column definitions all look solid. I think that this is ready to merge.
 
---
 
## What's Been Addressed
 
| Issue | Status |
|---|---|
| Duplicated `competitions` table + type exports | ✅ Removed |
| Bundled `admin/metrics` endpoint (FR-11) | ✅ Removed |
| `is_visible` defaulting to `true` | ✅ Fixed — now defaults to `false` |
| Redundant index on `learnworlds_user_id` | ✅ Removed |
| Trigger ordering (`DROP TRIGGER IF EXISTS` before `CREATE TRIGGER`) | ✅ Fixed |
| Missing newline at end of SQL file | ✅ Fixed |
 
---
 
## Summary
 
The schema is well-defined, the Drizzle schema matches the SQL, the trigger is correctly idempotent, and the RLS policy is correct for public read + `service_role` writes. This is ready to merge.
 
### PR Dependency & Merge Order
 
1. **This PR #90 (MD-05)** — `mentor_profiles` table in Supabase · *DB schema needed before any backend endpoint*
2. **#97 (IC-19)** — LearnWorlds ingestion foundation schema (`learnworlds_sync_runs`, `learnworlds_raw_payloads`, `learner_progress`)
3. **#98 (IC-11)** — LearnWorlds ingestion client (OAuth2 auth + fetch)
4. **#106 (MD-06)** — `POST /api/webhooks/learnworlds/mentor` endpoint ( branches off from this PR) 
5. **#108 (MD-01)** — LearnWorlds custom user fields (`cf_mentor_*`)
6. **#107 (MD-02)** — "Mentor Onboarding" Form Learning Activity

PR #78, opened 2026-03-03
# feat: Create GET /api/challenges endpoint with CORS support (IC-02, Issue #53)

sserurich:
# feat: Create GET /api/challenges endpoint with CORS support (IC-02, Issue #53 #79 )

## Description

Implement a public-facing API endpoint to fetch active challenges for integration with the LearnWorlds portal. The endpoint enables external platforms to display challenge cards to learners without requiring authentication.

---

### New Files

- `app/api/challenges/route.ts` — Public API endpoint for fetching challenges with CORS headers
- `tests/challenges-api.spec.ts` — Comprehensive Playwright test suite (11 tests)

### Modified Files

- `middleware.ts` — Excluded `/api/challenges` from auth middleware (public endpoint)
- `.env.local` — Added environment variables for CORS origin and signup URL configuration

---

#### Query Parameters

- `status` (optional): Filter challenges by status (`setup`, `open`, `active`, `completed`)
- Defaults to `open` if omitted or invalid

---

### CORS Support

- Handles `OPTIONS` preflight requests (204 status)
- Respects `LEARNWORLDS_ALLOWED_ORIGIN` environment variable
- Includes:
  - `Access-Control-Allow-Methods`
  - `Access-Control-Allow-Headers`
  - `Access-Control-Max-Age`
- Blocks CORS if origin does not match

---

### Response Format

```json
{
  "challenges": [
    {
      "id": "uuid",
      "title": "Challenge Name",
      "short_description": "Brief overview...",
      "cover_image_url": null,
      "challenge_type": "global",
      "tags": [],
      "prize_amount": null,
      "deadline": null,
      "teams_registered_count": 12,
      "participant_signup_url": "http://localhost:3000/participant/event/{id}"
    }
  ]
}
```

---

## Error Handling

- Returns `400` for invalid status parameters with descriptive error message
- Returns `500` for database/server errors
- Applies CORS headers to all responses

---

## Database Query

- Left joins `events` to count registered teams per challenge
- Groups by event ID and aggregates team count
- Orders results by creation date

---

#### Install dependencies

```bash
npm install
```

#### Start dev server (if not running)

```bash
npm run dev
```

#### Run all challenge API tests

```bash
npx playwright test tests/challenges-api.spec.ts
```

#### Run with specific browser

```bash
npx playwright test tests/challenges-api.spec.ts --project=chromium
```

---

#### Test basic endpoint

```bash
curl -s http://localhost:3000/api/challenges | jq .
```

#### Test with specific status

```bash
[...]

PR #77, opened 2026-02-27
feat: create supabase migration for challenges table (#52)

zsin520:
## **Changes**
- Added the challenges table to act as the public-facing bridge between internal events and the LearnWorlds display
- Created challenge_type and challenge_status ENUM types
- Established a foreign key relationship (event_id) linking to the events table
- Enabled Row Level Security (RLS) on the challenges table
- Configured RLS policies to allow Admins full management access (using check_user_role) and grant public read access for the LMS widget

## **Migration**
supabase/migrations/features/004_create_challenges_table.sql

## **Notes**
- Added database indexes on event_id and status to optimize API filtering and queries

PR #93, opened 2026-03-26
feat(IC-05): Auto-generate participant signup URL for competitions via trigger

### LINKED ISSUE #56: IC-05 feat: auto-generate participant_signup_url logic
Implement logic to automatically generate the `participant_signup_url` field to ensure consistency and eliminate manual data entry errors.

## Logic
Whenever a challenge is created or updated:
- **Format:** `https://judgeportal.com/participant/challenges/{event_id}`
- **Note:** The URL must use the `event_id`, not the challenge ID.

## Acceptance Criteria
- [x] URL is automatically populated upon record creation in the database.
- [x] Existing records update correctly if the linked `event_id` is changed.
- [x] Logic implemented via a Supabase Database Trigger or the backend service layer.

Thomas12556:
Adds a PostgreSQL trigger to the competitions table to automatically generate and maintain the participant_signup_url based on the associated event_id. This ensures the URL is always perfectly in sync with the event, eliminating the need to manually update the URL.

Changes
---

* New Trigger Function: Created set_competition_participant_signup_url() to build the URL format: https://judgeportal.com/participant/challenges/{event_id}.

* The trigger executes on insertions or updates.

* Data Backfill: DO block to backfill the URL for all existing records in the database.

Additional Notes
---
1. The trigger depends on the competitions table implemented in IC-01.
2. Tested using the following SQL code snippet:
```sql
-- Ensure changes are temporary
BEGIN; 

-- 1. Create test event A and test INSERTION
WITH new_event AS (
  INSERT INTO public.events (
    id,
    name,
    status,
    created_at,
    updated_at
  )
  VALUES (
    gen_random_uuid(),
    'Trigger Test Event',
    'setup',
    NOW(),
    NOW()
  )
  RETURNING id
)
INSERT INTO public.competitions (event_id)
SELECT id FROM new_event
RETURNING id, event_id, participant_signup_url;
-- Note: Copy the generated competition 'id' from the output above to use in step 3.

-- 2. Create test event B
INSERT INTO public.events (
  id,
  name,
  status,
  created_at,
  updated_at
)
VALUES (
  gen_random_uuid(),
  'Update Trigger Test Event B',
  'setup',
  NOW(),
  NOW()
)
RETURNING id;
-- Note: Copy the generated event 'id' from the output above to use in step 3.

-- 3. Test on UPDATE on event_id (manually update event A's event_id to event B's)
UPDATE public.competitions
SET event_id = '<PASTE_EVENT_B_UUID_HERE>' -- paste Event B uuid
WHERE id = '<PASTE_COMPETITION_ID_HERE>' -- paste competition id
RETURNING id, event_id, participant_signup_url;

-- Restore the changes made to the database
ROLLBACK;
```


Connor-Palmira (commented):
This PR is a solid implementation of URL generation logic. The manual testing that you provided is also really good at verifying the behavior of your code.

One thing I noticed in `supabase/migrations/features/005_auto-generate-participant-signup-url.sql` is that you hardcoded `https://judgeportal.com` which might create issues in local development and it might be a better idea to only store the relative path.

Overall, this is a really good and well-documented PR and you successfully implemented the URL generation through the database.

bofengb:
Hey , thanks for putting this together. The trigger-based approach for auto-generating `participant_signup_url` is a thoughtful way to enforce consistency at the database level.

After reviewing the diff alongside the other open PRs and the current state of `main`, I have a few items that need attention before this can move forward.

---

### 1. Accidental `git` File

The diff includes:
```diff
diff --git a/git b/git
new file mode 100644
```

This creates an empty file named `git` in the repo root — likely committed by accident. Please remove it.

---

### 2. Duplicated `competitions` Table and Migration

This PR includes the `competitions` table definition in `lib/db/schema.ts` and `supabase/migrations/features/003_competitions.sql`. Both of these were already merged into `main` via PR #77 (`feat/IC-01-challenges-table`).

Including them here will cause merge conflicts and potentially overwrite the already-merged version. Please **remove** the following from this PR:

- The `competitions` table definition and type exports from `lib/db/schema.ts`
- `supabase/migrations/features/003_competitions.sql`

Your migration file `005_auto-generate-participant-signup-url.sql` correctly lists `003_competitions.sql` as a dependency, so it just needs the duplicate removed — the dependency is already satisfied on `main`.

---

### 3. Hardcoded URL in Trigger

The trigger hardcodes:

```sql
new.participant_signup_url := 'https://judgeportal.com/participant/challenges/' || new.event_id::text;
```

A few concerns:
- The app is currently deployed at `judge-portal.vercel.app`, not `judgeportal.com`. If the domain changes again, we'd need a migration to update the trigger.
- PR #78 (`feature/api-challenges-cors`) generates this same URL in the **application layer** using an environment variable (`PUBLIC_PARTICIPANT_SIGNUP_BASE_URL`), which is more flexible and environment-aware.

**Recommendation:** Consider whether the trigger or the application layer should own this URL generation, Having both will lead to conflicts. If you keep the trigger, at least make the base URL configurable (e.g., via a Supabase Vault secret or a config table) rather than hardcoding it.

---

### 4. Test Snippet Committed

`supabase/snippets/Untitled query 561.sql` is a local test query that shouldn't be in the repo. Please remove it.

---

### 5. Supabase Config Files

`supabase/.gitignore` and `supabase/config.toml` are already being handled in PR #101 (`chore/supabase-local-dev-config`), which was specifically split out to keep feature branches scoped. Including them here duplicates that effort and creates merge conflicts. Please remove them from this PR.

---

### 6. `package-lock.json` Noise

There are significant `package-lock.json` changes with no corresponding `package.json` changes — likely from a different npm version. Please revert with:

```bash
git checkout main -- package-lock.json
```

---

### 7. `SECURITY DEFINER` on Trigger Function

The trigger function uses `SECURITY DEFINER`, which means it runs with the privileges of the function creator (typically the superuser). Since this trigger only modifies the row being inserted/updated in the same table, `SECURITY INVOKER` (the default) is sufficient and safer. Please remove the `SECURITY DEFINER` clause unless there's a specific RLS bypass requirement.

---

### 8. Silent Error Handling

```sql
EXCEPTION
  WHEN OTHERS THEN
    RAISE WARNING 'Error in set_competition_participant_signup_url for %', SQLERRM;
    RETURN new;
```

This catches ALL exceptions and silently continues. If there's a genuine error (e.g., a NULL `event_id` that shouldn't be NULL), the row will be inserted without a URL and the error will only appear in Postgres logs. Consider either letting the exception propagate or being more specific about which exceptions to catch.

---

### Summary

The core feature (a trigger to auto-populate the URL) is a valid approach, but this PR needs cleanup to remove duplicated schema/migration files, the accidental `git` file, and the test snippet. The hardcoded URL and overlap with PR #78's application-layer approach also need a decision on which layer owns URL generation.

Looking forward to the next iteration!

Thomas12556:
Thanks for the review @Connor-Palmira and @bofengb.

Cleanup
---

Accidental Files: Removed the root `/git file` and the local SQL test snippet `Untitled query 561.sql`.

Duplicates: Removed the competitions schema from `lib/db/schema.ts` and deleted the duplicate migration (003_competitions.sql), as these are already merged in main.

Configs: Reverted changes to `package-lock.json`, `supabase/.gitignore`, and `supabase/config.toml` to avoid conflicts with PR #101.

Refactor
---

Security: Changed the trigger function from SECURITY DEFINER to the default Invoker.

Error Handling: Removed the silent EXCEPTION block. The trigger now intentionally fails if the required configuration is missing, preventing rows from being inserted with broken URLs.

Dynamic URL: To avoid hardcoding, I refactored the trigger to dynamically retrieve the URL from `vault.decrypted_secrets` using the identifier `participant_signup_base_url`. 

Testing
---
Tested functionality in a local environment using the same test SQL code snippet given in the description of this PR under Additional Notes.
Initialized vault secret using:
```sql
SELECT vault.create_secret(
  'http://localhost:3000', 
  'participant_signup_base_url', 
  'Local dev base URL'
);
```

JBishopCode (changes requested):
# PR Review: IC-05 `feat/IC-05-auto-generate-participant-signup-url`

Went back through both previous review rounds and verified everything:

**From Connor's review:**
- ✅ Hardcoded domain — resolved via `vault.decrypted_secrets` lookup

**From Bofeng's review:**
- ✅ Accidental `git` file — removed
- ✅ Duplicated `competitions` schema and `003_competitions.sql` — removed
- ✅ Hardcoded URL in trigger — replaced with Vault lookup
- ✅ Test snippet `Untitled query 561.sql` — removed
- ✅ Supabase config files — reverted
- ✅ `package-lock.json` noise — reverted
- ✅ `SECURITY DEFINER` — dropped, now defaults to `SECURITY INVOKER`
- ✅ Silent `EXCEPTION WHEN OTHERS` — removed, errors now propagate properly

Also confirmed the migration file was renumbered to `006_` to avoid collision with the existing `005_` on main, and merge conflicts have been resolved.

The Vault-based approach is a clean solution to the hardcoded URL problem, and the stricter error handling is a solid improvement.

---

## One Blocker

The Playwright CI run failed on the migration step when pulling into main:

```
❌ auto-generate-participant-signup-url failed: Vault Error: Secret "participant_signup_base_url" is missing during backfill.
```

The `RAISE EXCEPTION` in the backfill `DO` block means any environment without `participant_signup_base_url` pre-seeded in Vault - including CI - will fail to apply this migration. This is going to block CI on every PR until it's resolved.

Once the backfill no longer hard-fails on a missing secret, this is good to go.

Thomas12556:
Thanks for the thorough review and catching the CI blocker. I've pushed an update to address the hard failure in the migration script.

Fixes
---
1. Changed `RAISE EXCEPTION` to `RAISE WARNING` inside the backfill DO block.
2. Changed feature name from 006_auto-generate-participant-signup-url.sql to 007_auto-generate-participant-signup-url.sql to resolve naming conflict with main branch.

sserurich:
> Thanks for the thorough review and catching the CI blocker. I've pushed an update to address the hard failure in the migration script.
> 
> ## Fixes
> 1. Changed `RAISE EXCEPTION` to `RAISE WARNING` inside the backfill DO block.
> 2. Changed feature name from 006_auto-generate-participant-signup-url.sql to 007_auto-generate-participant-signup-url.sql to resolve naming conflict with main branch.

Hi @Thomas12556 , thanks for addressing the reviews feedback. All of Connor’s and bofengb’s requested changes appear to be addressed. JBishopCode’s blocker is not clearly fully resolved, because while the backfill no longer hard-fails, the trigger still hard-fails when the secret is absent. Will merge when this is also addressed.

The backfill DO $$ ... $$ block now uses RAISE WARNING + RETURN; when the secret is missing, which addresses the backfill failure mode.
However, the trigger function itself still does RAISE EXCEPTION if the secret is missing:
```SQL
IF base_url IS NULL OR base_url = '' THEN
  RAISE EXCEPTION 'Vault Error: Secret "participant_signup_base_url" is missing.';
END IF; 
```
That means any INSERT/UPDATE to competitions.event_id in environments without the secret will still fail. If CI (Playwright) creates/updates competitions rows during tests, CI can still be broken even after the backfill change.

Thomas12556:
Thank you Richard for the feedback. I have addressed the blocker in the trigger function with the following:
```sql
  IF base_url IS NULL OR base_url = '' THEN
    RAISE WARNING 'Vault Error: Secret "participant_signup_base_url" is missing. Skipping URL generation.';
    RETURN NEW;
  END IF;
```

sserurich:
> Thank you Richard for the feedback. I have addressed the blocker in the trigger function with the following:
> 
> ```sql
>   IF base_url IS NULL OR base_url = '' THEN
>     RAISE WARNING 'Vault Error: Secret "participant_signup_base_url" is missing. Skipping URL generation.';
>     RETURN NEW;
>   END IF;
> ```

Thanks @Thomas12556 .

sserurich (approved):
Looks good now after implementing all the feedback I shared and what others had shared.

PR #101, opened 2026-03-28
chore: add supabase local dev config and .gitignore

nsoliven:
## Summary
- Add `supabase/.gitignore` to exclude local dev artifacts (`.branches`, `.temp`, env files)
- Add `supabase/config.toml` for local Supabase development

Split out from #84 per @bofengb's [review feedback](https://github.com/Freddie-Pike/judge-portal/pull/84#issuecomment) — keeping feature branches scoped to their feature.

## Question for team
Should we commit `config.toml` so anyone can `supabase start` locally without needing remote credentials? Or should this stay local-only (gitignored)?

Right now running/testing locally requires a remote Supabase instance which seems unnecessary when `supabase start` gives you everything out of the box.

## Test plan
- No functional changes, config and gitignore only

### DIFF
--- a/app/api/admin/metrics/route.ts
+++ b/app/api/admin/metrics/route.ts
@@ -15,23 +15,20 @@ import { eq } from 'drizzle-orm';
 export async function GET() {
   try {
     const user = await getUserFromSession();
-    
-  // Only admins can access this endpoint
+
+    // Only admins can access this endpoint
     if (!user || user.role !== 'admin') {
       return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
     }
 
     // Query total learners (participants)
-    const learners = await db
-      .select()
-      .from(users)
-      .where(eq(users.role, 'participant'));
-    
-// TODO: Replace placeholder metrics with real analytics queries from the database
+    const learners = await db.select().from(users).where(eq(users.role, 'participant'));
+
+    // TODO: Replace placeholder metrics with real analytics queries from the database
     const metrics = {
       totalRecommendationRequests: 0,
       mostFrequentlyRecommendedItems: [],
-      totalLearnersServed: learners.length, 
+      totalLearnersServed: learners.length,
       averageResponseTime: 0,
     };
 
--- a/supabase/.gitignore
+++ b/supabase/.gitignore
@@ -1,0 +1,8 @@
+# Local dev artifacts
+.branches
+.temp
+
+# Environment files
+.env.keys
+.env.local
+.env.*.local
--- a/supabase/config.toml
+++ b/supabase/config.toml
@@ -1,0 +1,384 @@
+# For detailed configuration reference documentation, visit:
+# https://supabase.com/docs/guides/local-development/cli/config
+# A string used to distinguish different Supabase projects on the same host. Defaults to the
+# working directory name when running `supabase init`.
+project_id = "judge-portal"
+
+[api]
+enabled = true
+# Port to use for the API URL.
+port = 54321
+# Schemas to expose in your API. Tables, views and stored procedures in this schema will get API
+# endpoints. `public` and `graphql_public` schemas are included by default.
+schemas = ["public", "graphql_public"]
+# Extra schemas to add to the search_path of every request.
+extra_search_path = ["public", "extensions"]
+# The maximum number of rows returns from a view, table, or stored procedure. Limits payload size
+# for accidental or malicious requests.
+max_rows = 1000
+
+[api.tls]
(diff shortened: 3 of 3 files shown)


bofengb:
# PR Review: `chore/supabase-local-dev-config`

Hi, after looking at how our migrations and DB scripts actually work, I don't think this PR fits our project's architecture, because our repo has deliberately chosen **not** to use the Supabase CLI's migration tooling, and adopting `config.toml` pulls in half of a parallel tooling layer that doesn't connect to the system we actually have.

---

## 1. The Goal

This PR aims to let contributors run `supabase start` locally to spin up the full Supabase stack (Postgres, Auth, Storage, Studio, Inbucket) without needing access to a remote project. The PR body asks the right question: _"Should we commit `config.toml` so anyone can `supabase start` locally without needing remote credentials? Or should this stay local-only (gitignored)?"_.

---

## 2. The Core Problem: Our Migration System Bypasses the Supabase CLI

This is the bigger architectural reason I don't think `config.toml` belongs here. Our repo has made a deliberate choice to manage migrations **outside** the Supabase CLI:

| Piece | What we have | What `supabase start` expects |
|------|-------------|-----------------|
| Migration location | `supabase/migrations/features/NNN_name.sql` + a top-level `consolidated_setup.sql` | `supabase/migrations/<timestamp>_name.sql` at the top level |
| Migration naming | Custom numeric prefix (`001_invite-link.sql` … `005_learnworlds-ingestion-foundation.sql`) | Timestamped (`20260101000000_name.sql`) |
| How migrations get applied | `scripts/consolidated-setup.ts` — connects directly via the `postgres` npm package using `DATABASE_URL` and runs SQL via a raw client | `supabase db push` / `supabase db reset` via the CLI |
| Schema push | `drizzle-kit push` (`npm run db:push`) — diffs Drizzle schema against DB | `supabase db push` |
| Feature migrations | `scripts/apply-feature-migration.ts` — custom applier for the `features/` subdirectory | Handled by CLI migration system |

And `config.toml` line 58–61 (`[db.migrations] schema_paths = []`) confirms it — schema discovery is empty, so even if you wanted the CLI to see our SQL, it wouldn't.

**What this means in practice:** if someone checks out this branch and runs `supabase start`, the CLI spins up a blank Postgres. None of our migrations apply. To get a usable schema, they'd still run `npm run db:setup` pointed at `postgresql://postgres:postgres@127.0.0.1:54322/postgres`. At that point, the CLI's local stack is acting as a generic Postgres container. We'd get an equivalent result from a 3-line `docker-compose.yml` without any of the CLI overhead.

So `config.toml` only buys us the Auth/Storage/Studio emulators and port standardization, which is a narrow slice of value for 384 lines of committed config that drifts from upstream Supabase templates.

## 3. Secondary Issues

### 3.1 The second commit is a no-op against current `main`

The second commit (`d817bd` — "chore: format metrics route") reformats `app/api/admin/metrics/route.ts`, but `main` is already in that exact formatted state #114 ("Fix formatting for Lint&Format error") landed this upstream. GitHub still reports `MERGEABLE`, but the commit is stale and makes the diff confusing on a chore PR.

**Fix:** rebase onto latest `main` and drop the formatting commit.

### 3.2 The `.gitignore` partially duplicates the root

Our root `.gitignore` already has:

```
.env*.local
.env
```

…which covers 3 of the 5 environment entries in `supabase/.gitignore`. The net-new rules are really just:

```
.branches
.temp
```

These two are worth keeping regardless of the strategic decision. They're cheap insurance against accidental commits if anyone experiments with `supabase start`. They could also just live in root `.gitignore` as `supabase/.branches` and `supabase/.temp`, removing the need for a nested `.gitignore` file entirely.

---

## 4. Recommendation

**Close the PR**. The reasoning, summarized:

1. Our migration system deliberately bypasses the Supabase CLI (custom SQL locations, custom naming, direct `postgres` client, Drizzle push). `config.toml` doesn't plug into any of it.
2. Adopting `supabase start` as a supported workflow would require either (a) rewriting our migrations into the CLI's timestamp format and migrating the setup scripts to `supabase db reset`, or (b) accepting that the CLI's local stack is just a dumb Postgres container for us. Option (a) is a real architectural decision that deserves an issue/RFC, not a chore PR. Option (b) has no value proposition that beats plain Docker.

If the team does want to revisit local-first dev later, the right starting point is a design discussion, not a config file.

sserurich:
@nsoliven , seems you need to close this PR following recommendation from @bofengb

sserurich:
@nsoliven , I have closed this PR following the recommendation by @bofengb.

Issue #102, opened 2026-03-28
IC-20: Standardize API Error Handling


zsin520:
### Description
FR-14 Basic Error Handling - The system shall validate API inputs and return appropriate HTTP status codes (400, 404, 500) with structured JSON error messages.

Currently, our API routes handle errors inconsistently. Various endpoints return different JSON shapes (e.g., { error: "..." } vs { errorType: "..." }) and sometimes use non-standard HTTP status codes.

To comply with Functional Requirement 14 (FR-14), we need a centralized utility to guarantee that all endpoints return the exact structured JSON format required by the specification.

### Tasks:
1. Create a reusable error utility at lib/utils/api-errors.ts.
2. Ensure the utility enforces the exact FR-14 JSON schema: { "error_code": string, "error_message": string }.
3. Ensure the utility restricts status codes to 400, 404, and 500.
4. Refactor the current active feature routes (MD-06, FR-9) to use this utility.

### Acceptance Criteria:
- [ ] sendApiError utility exists and is fully typed.
- [ ] Calling the utility produces a strictly formatted JSON response.

PR #103, opened 2026-03-28
IC-20: Standardize API Error Handling

zsin520:
## feat: implement standardized API error utility for FR-14
FR-14 Basic Error Handling - The system shall validate API inputs and return appropriate HTTP status codes (400, 404, 500) with structured JSON error messages.

Currently, our API routes handle errors inconsistently. Various endpoints return different JSON shapes (e.g., { error: "..." } vs { errorType: "..." }) and sometimes use non-standard HTTP status codes.

To comply with Functional Requirement 14 (FR-14), we need a centralized utility to guarantee that all endpoints return the exact structured JSON format required by the specification.

## Changes made
- Created Utility: Added lib/utils/api-errors.ts featuring the sendApiError() function.
- Type Safety: Restricted HTTP status codes to 400 | 404 | 500 via TypeScript union types to enforce the MVP spec.

## Notes
- I did not refactor any existing API routes in this PR. Once this PR is merged, if you are creating a new route.ts, please use the new sendApiError() utility.

## Questions
- Are there any recommendations for the best approach to implement sendApiError() utility in existing route.ts files ? 
- I restricted the HTTP status types strictly to 400 | 404 | 500 to match the FR-14 specification. However, we may need to throw 401 Unauthorized or 403 Forbidden errors through this utility soon? Let me know if you think we should expand this union type.


app/api/admin/metrics/route.ts
```diff
@@ -15,23 +15,20 @@ import { eq } from 'drizzle-orm';
...
     }
 
     // Query total learners (participants)
-    const learners = await db
-      .select()
-      .from(users)
-      .where(eq(users.role, 'participant'));
-    
-// TODO: Replace placeholder metrics with real analytics queries from the database
+    const learners = await db.select().from(users).where(eq(users.role, 'participant'));
```
8bit-Apollo: Displaying this on multiple lines as before is slightly better for readability

8bit-Apollo (commented):
Everything looks good. Code is readable, well documented, and appears functional. The one fix I spotted was extremely minor. Good job!

 In response to your question, I think adding 401 and 403 type errors would be a prudent expansion to this requirement, but let's discuss that at our team meeting.

bofengb:
# PR Review: `util/api-errors`

Hi, thanks for kicking off the FR-14 standardization work. I reviewed this alongside the related issue (#102) and the surrounding API routes, and I have a few suggestions before we merge.

---

## 1. The Goal

This PR closes out issue #102 (IC-20) which asks for:

1. A reusable helper at `lib/utils/api-errors.ts`
2. Enforcement of the FR-14 response shape `{ error_code, error_message }`
3. Restricting status codes to `400 | 404 | 500`
4. Refactoring the current active feature routes (MD-06, FR-9) to use the helper

The intent is to stop the current drift where some endpoints return `{ error: "..." }`, a subset of `/api/judge/*` routes return `{ error, errorType }`, and status codes are inconsistent.

---

The union `400 | 404 | 500` matches the FR-14 spec wording, but I see:

- other responses using `401 / 403 / 409 / 422`
- The route in this very PR returns `401` for the auth guard

That means `sendApiError()` literally cannot be called from the file the PR modifies, and can't be adopted by any auth-guarded route (which is almost all of them). The helper would end up applicable only to a handful of validation-error paths.

**Fix:** I'd expand the union to cover what the codebase actually uses:

```ts
export type ApiErrorStatus = 400 | 401 | 403 | 404 | 409 | 422 | 500;

export function sendApiError(status: ApiErrorStatus, errorCode: string, errorMessage: string) {
  return NextResponse.json({ error_code: errorCode, error_message: errorMessage }, { status });
}
```

---

This PR has no hard code dependencies on other open PRs, but it does have a *policy* dependency: the error-shape decision should be made before this merges, because whatever shape lands here will be copied into every follow-up route refactor.

Issue #61, opened 2026-02-17
IC-10 test: qa challenge flow with learner accounts


SaifAhmed1902:
Verify the end-to-end flow using actual learner accounts to ensure data passing works in the wild.

### Test Scenarios
1. **New User:** Click Apply -> Redirect -> Verify Judge Portal signup form is pre-filled.
2. **Returning User:** Click Apply -> Redirect -> Verify Judge Portal login form is pre-filled.
3. **Guest (Logged out):** Verify graceful degradation (e.g., Liquid variables are empty; ensure link doesn't break).

### Tasks
- [x] Test with Admin account
- [x] Test with standard Learner account
- [x] Test on Mobile browser

### Acceptance Criteria
- [x] Redirects work consistently
- [x] User data is accurately passed in URL params

lolbro2k:
Added test account under judge, learning and admin.

New Users/Sign up:
URL params only update the liquid variable, Role, of the account. Redirects work consistently on both new User and Mobile.

Returning Users/Login:
Sign up works as expected expect Only the role is passed into URL params. 

Guest/Logged out:
Degradation of roles for the URL params works.

tested on Mobile, Admin, and Participant accounts.
Redirect consistent, but does not pass User Data in URL params

##Passing
- redirects work consistently with learner, admin, super_admin, and on mobile browser
- role-based User Data is passed in URL params
- graceful degradation when logged out
- User Roles are passed in URL params, LearnWorlds is similar once logged in, so consistent URL params passing

PR #105, opened 2026-03-29
IC-10 QA challenge flow test

lolbro2k:
#61 Verify the end-to-end flow using actual learner accounts to ensure data passing works in the wild.

This is an empty PR to document the testing of QA challenge flow test, no code was changed only documentation of the learner accounts in the wild

Added test account under judge, learning and admin.

New Users/Sign up:
URL params only update the liquid variable, Role, of the account. Redirects work consistently on both new User and Mobile.

Returning Users/Login:
Sign up works as expected expect Only the role is passed into URL params. 

Guest/Logged out:
Degradation of roles for the URL params works.

tested on Mobile, Admin, and Participant accounts.
Ran server on local machine and local network then connected through IP on IOS mobile device.
Re ran the previous test on Mobile to ensure mobile degradation of variable and prefill sign up works

Redirect consistently and URL role-based data is passed in URL params, like Learnworld side accepts

##Passing
- redirects work consistently with learner, admin, super_admin, and on mobile browser
- role-based User Data is passed in URL params
- graceful degradation when logged out
- User Roles are passed in URL params, LearnWorlds is similar once logged in, so consistent URL params passing


sserurich:
Thanks @lolbro2k for testing this end to end flow. Great to know it works. I have seen @saif1902  closed the issue after your feedback. so am closing this PR.

PR #106, opened 2026-03-29
feat(MD-06): webhook receiver for mentor registration

### LINKED ISSUE #67: MD-06 feat: create webhook receiver for mentor registration
Develop an API endpoint in the Judge Portal to receive real-time data from LearnWorlds when a mentor submits their application (this will be public form created through LearnWorlds).

### Context
This endpoint will be the "listener" for the LearnWorlds webhook.

### Tasks
- [ ] Create endpoint `POST /api/webhooks/learnworlds/mentor`
- [ ] Validate webhook signature (if LearnWorlds provides a secret) or basic auth header
- [ ] Parse payload to extract Custom User Fields (`cf_mentor_title`, `cf_mentor_bio`, etc.)
- [ ] Transform/Map LearnWorlds fields to Supabase `mentor_profiles` columns
- [ ] Perform UPSERT operation (insert if new, update if exists) based on email or `learnworlds_user_id`

### Acceptance Criteria
- [ ] Posting dummy JSON payload to endpoint creates a new record in `mentor_profiles`
- [ ] Updates to existing email update the record instead of creating duplicates

nsoliven:
Implements POST /api/webhooks/learnworlds/mentor (issue #67).

When a mentor submits their application on LearnWorlds, this endpoint
receives the webhook, maps the custom fields to our mentor_profiles
table, and upserts the record.

- Webhook secret validation with constant-time comparison
- Maps cf_mentor_* fields to mentor_profiles columns
- Atomic upsert via onConflictDoUpdate (no race conditions)
- New profiles default to isVisible: false (pending admin approval)
- 5 Playwright tests covering auth, validation, and upsert

Depends on #90 (MD-05 mentor_profiles table).
Closes #67.

### DIFF
--- a/app/api/webhooks/learnworlds/mentor/route.ts
+++ b/app/api/webhooks/learnworlds/mentor/route.ts
@@ -1,0 +1,78 @@
+import { NextRequest, NextResponse } from 'next/server';
+import { timingSafeEqual } from 'crypto';
+import { db } from '@/lib/db';
+import { mentorProfiles } from '@/lib/db/schema';
+
+// LearnWorlds custom field names → mentor_profiles columns
+const FIELD_MAP: Record<string, string> = {
+  cf_mentor_title: 'title',
+  cf_mentor_org: 'organization',
+  cf_mentor_bio: 'bio',
+  cf_mentor_linkedin: 'linkedinUrl',
+  cf_mentor_calendly: 'calendlyUrl',
+  cf_mentor_photo: 'photoUrl',
+};
+
+/**
+ * POST /api/webhooks/learnworlds/mentor
+ * Receives webhook from LearnWorlds when a mentor submits their application.
+ * Upserts into mentor_profiles based on learnworlds_user_id.
+ */
+export async function POST(request: NextRequest) {
+  try {
+    const webhookSecret = process.env.LEARNWORLDS_WEBHOOK_SECRET;
+    if (!webhookSecret) {
+      console.error('LEARNWORLDS_WEBHOOK_SECRET is not configured');
+      return NextResponse.json({ error: 'Server misconfigured' }, { status: 500 });
+    }
+
+    // Validate webhook secret with constant-time comparison
+    const signature = request.headers.get('x-webhook-secret') ?? '';
+    const sigBuf = Buffer.from(signature);
+    const secretBuf = Buffer.from(webhookSecret);
+    if (sigBuf.length !== secretBuf.length || !timingSafeEqual(sigBuf, secretBuf)) {
+      return NextResponse.json({ error: 'Invalid signature' }, { status: 401 });
+    }
+
+    const payload = await request.json();
+
+    const userId = payload.user_id || payload.id;
+    const fullName = payload.name || payload.full_name || 'Unknown';
+
+    if (!userId) {
+      return NextResponse.json({ error: 'Missing user_id' }, { status: 400 });
+    }
+
+    // Map LearnWorlds custom fields to our column values
+    const mapped: Record<string, string | null> = {};
+    for (const [lwField, dbColumn] of Object.entries(FIELD_MAP)) {
+      if (payload[lwField] !== undefined) {
+        mapped[dbColumn] = payload[lwField];
+      }
+    }
+
+    // Atomic upsert — insert new profile or update existing by learnworlds_user_id
+    await db
+      .insert(mentorProfiles)
+      .values({
+        learnworldsUserId: userId,
+        fullName,
+        title: mapped.title ?? null,
+        organization: mapped.organization ?? null,
+        bio: mapped.bio ?? null,
+        linkedinUrl: mapped.linkedinUrl ?? null,
+        calendlyUrl: mapped.calendlyUrl ?? null,
+        photoUrl: mapped.photoUrl ?? null,
+        isVisible: false,
+      })
+      .onConflictDoUpdate({
+        target: mentorProfiles.learnworldsUserId,
+        set: { fullName, ...mapped },
+      });
+
+    return NextResponse.json({ status: 'ok' });
+  } catch (error) {
+    console.error('Webhook error:', error);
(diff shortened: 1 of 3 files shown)


Thomas12556 (approved):
This is a great implementation of the core webhook infrastructure that meets all the criteria of the feature. 

Notable Implementation Choices
--- 
- Using .onConflictDoUpdate for the upsert ensures that if LearnWorlds sends duplicate webhooks, the database would remain consistent without creating redundant rows. 
- Defaulting isVisible to false ensures that the Admins are not potentially swarmed with spam of the mentors.
- The playwright tests provide code coverage for payload validation and adds security.

Overall, I am in approval of merging this PR as it provides a secure foundation which my subsequent [PR #111](https://github.com/Freddie-Pike/judge-portal/pull/111) builds upon by adding the `role_mentor` tag filter.

bofengb:
# PR Review: `feat/MD-06-mentor-webhook`

Hi, thanks for getting this webhook up. It's a solid foundation for the whole mentor-sync pipeline. I reviewed this alongside #90 (MD-05, now merged) and #111 (MD-08, the tag-filter follow-up) so the comments here account for what's already landed on main and what's coming next.

---

## 1. The Goal

This PR implements `POST /api/webhooks/learnworlds/mentor` (issue #67) so that when a mentor submits their application through LearnWorlds, the payload is verified, mapped from `cf_mentor_*` custom fields onto `mentor_profiles` columns, and upserted atomically keyed on `learnworlds_user_id`. New profiles default to `isVisible: false` so admins approve them before they show up publicly.

---

### 2.1 Merge conflict with `main` — duplicate schema and migration

Since this branch was cut, #90 (MD-05) was merged on 2026-04-15, and it already added the `mentor_profiles` table to both `lib/db/schema.ts` and `supabase/migrations/features/004_mentor_profiles.sql`. This PR re-adds those exact artifacts, which is why GitHub is showing `CONFLICTING` / `DIRTY`.

**Fix:** Rebase onto `main` and drop the duplicate changes from this PR so only the webhook-specific files remain.

After rebase, the only files in the diff should be:

- `app/api/webhooks/learnworlds/mentor/route.ts` (new)
- `tests/mentor-webhook.spec.ts` (new)

### 2.2 `/api/webhooks/*` is not in the middleware's public allowlist

The root middleware matcher in `middleware.ts` excludes only `api/challenges`, so the webhook path does run through `updateSession` in `lib/supabase/middleware.ts:51-61`. That block lists the unauthenticated allowlist as `/`, `/auth`, `/invite`, `/api/invite`, `/api/organizations/public` — no webhook path. When LearnWorlds posts here with no Supabase session cookie, the request gets 302-redirected to `/auth/login` before it reaches the handler, so the webhook never actually runs in production.

**Fix:** Add `/api/webhooks` to the unauthenticated allowlist:

```ts
// lib/supabase/middleware.ts (around line 53)
if (
  pathname === '/' ||
  pathname.startsWith('/auth') ||
  pathname.startsWith('/invite') ||
  pathname.startsWith('/api/invite') ||
  pathname.startsWith('/api/organizations/public') ||
  pathname.startsWith('/api/webhooks')
) {
  return supabaseResponse;
}
```

I'd also suggest adding a Playwright test that posts without a Supabase session cookie and asserts we get `200`/`401`/`400` from the handler — not a `302` from middleware. That would have caught this.

### 2.3 Out-of-scope deletion of `app/api/admin/metrics/route.ts`

The diff removes `app/api/admin/metrics/route.ts` (43 lines), which is the FR-11 / IC-17 admin analytics endpoint — completely unrelated to MD-06. Tracing the history, it looks like this slipped in from when #90 briefly carried that file before cleaning up in commit `cd3ec08`. The file is live on `main` and other in-flight PRs depend on it.

**Fix:** Restore the file during the rebase. After the rebase in 2.1, this should fall out naturally if the schema/migration hunks are dropped, but double-check the diff before pushing.

### 2.4 `isVisible` default drift in this PR

This PR's `004_mentor_profiles.sql` has `is_visible BOOLEAN DEFAULT TRUE NOT NULL` and the Drizzle schema has `.default(true)`, but the PR description says new profiles default to `false`. The handler hardcodes `isVisible: false` on insert so the webhook path is safe, but any future direct insert (seeds, admin tools) would default to visible, which defeats the approval gate.

Good news: `main` already has `default(false)` correctly (MD-05 corrected it), so once 2.1's rebase drops the duplicate schema/migration hunks, this auto-resolves. Just worth confirming after the rebase.

---

Based on the mentor-discovery chain, suggested merge order:

1. **#90 MD-05 mentor_profiles table** — already merged ✅
2. **#106 MD-06 webhook receiver** — this PR (after rebase + fixes above)
3. **#111 MD-08 LearnWorlds webhook config / tag filter** — builds on #106
4. **#160 MD-11 admin mentor approval API** — uses the profiles this webhook creates
5. **#161 MD-12 admin mentor management UI** — targets #160's branch

#109 (MD-07 public mentors API) is parallel to this. It only depends on #90, so it can merge independently.

AngelVargas1009 (commented):
The implementation looks like it addresses the mentor registration webhook feature well, especially with the webhook receiver, validation/security considerations, and mentor profile creation flow.

However, I don’t think this PR is ready to merge yet because there are merge conflicts that still need to be resolved, including conflicts in the schema and migration files. Since those files affect the database structure, I think they should be reviewed carefully before merging to avoid breaking the mentor profile setup or duplicating migration changes.

Once the conflicts are resolved and the checks stay green, this should be much closer to merge-ready.

nsoliven:
Addressed the review feedback: rebased onto main, removed the stale schema/migration and unrelated metrics diff, added /api/webhooks to the unauthenticated middleware allowlist, and added a regression test so webhook POSTs do not redirect to login.

sserurich (approved):
Thanks @nsoliven , this looks good and feedback has been implemented. The only thing I have added is to ensure the documentation for this API is created. I have created a separate issue for this  #195 .

PR #95, opened 2026-03-27
Feat: MD-09 -- Build a Find a Mentor Widget

Connor-Palmira:
Adds a Find a Mentor widget that displays all mentors from the /api/mentors endpoint and displays it. This widget has mentor information, their LinkedIn button, and a Book Session button. While the data is loading, a Loading State UI is displayed. The mentor widget collapses from a 3-card wide grid to 1-card wide on mobile.

Changes
- Adds the Mentor Widget to components
- Adds the mentors pages to the parcipant portal
- Adds css to /app/globals.css for mentor cards

Additional Notes:

- This pull request relies on MD-07
- This was tested using the following mock data put into /api/mentors/route.ts

```
import { NextResponse } from 'next/server';

export async function GET() {
  const mockMentors = [
    {
      // Test for default avatar
      id: '1',
      learnworld_user_id: '1',
      full_name: 'full name',
      title: 'title',
      organization: 'a',
      bio: 'a',
      linkedin_url: 'https://linkedin.com',
      calendly_url: 'https://calendly.com',
      photo_url: null,
      tags: ['a', 'b', 'c'],
      is_visible: true,
    },
    {
      // Test for fallback for default avatar
      id: '2',
      learnworld_user_id: '2',
      full_name: 'Default fallack',
      title: 'Test',
      organization: 'org',
      bio: 'a',
      linkedin_url: 'https://linkedin.com',
      calendly_url: 'https://calendly.com',
      photo_url: null,
      tags: ['a', 'b', 'c'],
      is_visible: true,
    },
    {
      // Test for long name, title, organization truncation
      // Test for tag wrapping for many tags
      // Test for image loading
      id: '3',
      learnworld_user_id: '3',
      full_name: 'Long Name Long Name Long Name Long Name',
      title: 'Long Title Long Title Long Title Long Title',
      organization: 'Long Organization Long Organization Long Organization',
      bio: 'bio',
      linkedin_url: 'https://linkedin.com',
      calendly_url: 'https://calendly.com',
      photo_url: 'https://thumbs.dreamstime.com/b/happy-man-thumbs-up-sign-full-length-portrait-white-background-showing-31416426.jpg',
      tags: ['tag1', 'tag2', 'tag3', 'tag4', 'tag5', 'tag6', 'tag7', 'tag8', 'tag9', 'tag10'],
      is_visible: true,
    },
    {
      // Test for visibility
      id: '4',
      learnworld_user_id: '4',
      full_name: 'Hidden',
      title: 'Hidden',
[...]

PR #107, opened 2026-03-29
MD-02: Adds mentor application form component and page

### LINKED ISSUE #63: MD-02 feat: build mentor application form activity
Create the user-facing interface for potential mentors to submit their details.

### Implementation
Create as a **Form Learning Activity** within a "Become a Mentor" standalone page. 

### Tasks
- [ ] Create a new Page: "Mentor Onboarding"
- [ ] Add Learning Activity: "Form / Survey"
- [ ] Add questions corresponding to the fields created in `MD-01`
- [ ] **Critical:** Map each question's answer to the specific Custom User Field (e.g., "Job Title" -> `cf_mentor_title`)
- [ ] Configure "Areas of Expertise" as a Multi-select question

### Acceptance Criteria
- [ ] Form is accessible to logged-in users
- [ ] Submitting the form updates the user's profile data in the backend

Connor-Palmira:
This PR implements the Mentor Onboarding form for the participant portal. It provides an interface for users to submit their professional details that are required for the mentor status.

Changes
- New page (app/participant/mentor-onboarding/page.tsx) that loads the Mentor Onboarding form
- New component (component/mentor-onboarding-form.tsx) that allows the users to upload their professional information to become a mentor. It adds questions corresponding to MD-01 and maps the answers to the custom user fields mentioned in MD-01.
- Success Notification that appears when submitted the Mentor Onboarding form before redirecting the Participant Portal

Additional Notes
- The "Areas of Expertise" multi-select question currently holds mock answers
- The information provided by the user is uploaded into their metadata

Closes #63

### DIFF
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -1,0 +1,15 @@
+// Placeholder for testing
+import { NextResponse } from 'next/server';
+
+export async function POST(req: Request) {
+  let body: any = null;
+  try {
+    body = await req.json();
+  } catch {
+    body = await req.text();
+  }
+
+  console.log(body);
+
+  return NextResponse.json({ ok: true });
+}
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -1,15 +1,13 @@
 import { NextResponse } from 'next/server';
-import { authServer } from '@/lib/auth';
 import { db } from '@/lib/db';
 import { mentorProfiles } from '@/lib/db/schema';
 
 export async function POST(req: Request) {
   try {
-    const user = await authServer.requireParticipant();
     const body = await req.json();
 
     const values = {
-      learnworldsUserId: user.id,
+      learnworldsUserId: null, // Needs updating when learnworlds user id is fetchable
       fullName: body.cf_mentor_name,
       title: body.cf_mentor_title,
       organization: body.cf_mentor_org,
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -1,15 +1,49 @@
-// Placeholder for testing
 import { NextResponse } from 'next/server';
+import { authServer } from '@/lib/auth';
+import { db } from '@/lib/db';
+import { mentorProfiles } from '@/lib/db/schema';
 
 export async function POST(req: Request) {
-  let body: any = null;
   try {
-    body = await req.json();
-  } catch {
-    body = await req.text();
+    const user = await authServer.requireParticipant();
+    const body = await req.json();
+
+    const values = {
+      learnworldsUserId: user.id,
+      fullName: body.cf_mentor_name,
+      title: body.cf_mentor_title,
+      organization: body.cf_mentor_org,
+      bio: body.cf_mentor_bio,
+      linkedinUrl: body.cf_mentor_linkedin,
+      calendlyUrl: body.cf_mentor_calendly,
+      photoUrl: body.cf_mentor_photo,
+      tags: body.cf_mentor_expertise ?? [],
+      isVisible: false,
+    };
+
+    await db
+      .insert(mentorProfiles)
+      .values(values)
+      .onConflictDoUpdate({
+        target: mentorProfiles.learnworldsUserId,
+        set: {
+          fullName: values.fullName,
+          title: values.title,
+          organization: values.organization,
+          bio: values.bio,
+          linkedinUrl: values.linkedinUrl,
+          calendlyUrl: values.calendlyUrl,
+          photoUrl: values.photoUrl,
+          tags: values.tags,
+        },
+      });
+
+    return NextResponse.json({ ok: true });
+  } catch (error: any) {
+    console.error('[MENTOR_APPLICATION_POST]', error);
+    return NextResponse.json(
+      { error: error.message || 'Internal Server Error' },
+      { status: error.status || 500 }
+    );
   }
-
-  console.log(body);
-
-  return NextResponse.json({ ok: true });
 }
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -39,11 +39,11 @@ export async function POST(req: Request) {
       });
 
     return NextResponse.json({ ok: true });
-  } catch (error: any) {
+  } catch (error) {
     console.error('[MENTOR_APPLICATION_POST]', error);
     return NextResponse.json(
-      { error: error.message || 'Internal Server Error' },
-      { status: error.status || 500 }
+      { error: 'Internal Server Error' }, 
+      { status: 500 }
     );
   }
 }
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -1,13 +1,14 @@
 import { NextResponse } from 'next/server';
 import { db } from '@/lib/db';
 import { mentorProfiles } from '@/lib/db/schema';
+import { sql } from 'drizzle-orm';
 
 export async function POST(req: Request) {
   try {
     const body = await req.json();
 
     const values = {
-      learnworldsUserId: null, // Needs updating when learnworlds user id is fetchable
+      learnworldsUserId: sql`null`,
       fullName: body.cf_mentor_name,
       title: body.cf_mentor_title,
       organization: body.cf_mentor_org,
--- a/app/api/mentor-application/route.ts
+++ b/app/api/mentor-application/route.ts
@@ -40,10 +40,7 @@ export async function POST(req: Request) {
 
     return NextResponse.json({ ok: true });
   } catch (error) {
-    console.error('[MENTOR_APPLICATION_POST]', error);
-    return NextResponse.json(
-      { error: 'Internal Server Error' }, 
-      { status: 500 }
-    );
+    console.error('Error during submission: ', error);
+    return NextResponse.json({ error: 'Internal Server Error' }, { status: 500 });
   }
 }
--- a/app/participant/mentor-onboarding/page.tsx
+++ b/app/participant/mentor-onboarding/page.tsx
@@ -1,0 +1,20 @@
+import { createClient } from '@/lib/supabase/server';
+import { redirect } from 'next/navigation';
+import { MentorOnboardingForm } from '@/components/mentor-onboarding-form';
+
+export default async function MentorOnboardingPage() {
+  const supabase = await createClient();
+  const {
+    data: { user },
+  } = await supabase.auth.getUser();
+
+  if (!user) {
+    redirect('/auth/login');
+  }
+
+  return (
+    <div className="container flex items-center justify-center min-h-[calc(100vh-64px)] py-10">
+      <MentorOnboardingForm />
+    </div>
+  );
+}
--- a/components/mentor-onboarding-form.tsx
+++ b/components/mentor-onboarding-form.tsx
@@ -3,7 +3,6 @@
 import { useState } from 'react';
 import { useRouter } from 'next/navigation';
 import { cn } from '@/lib/utils';
-import { createClient } from '@/lib/supabase/client';
 import { Button } from '@/components/ui/button';
 import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/components/ui/card';
 import { Input } from '@/components/ui/input';
@@ -18,6 +17,7 @@ export function MentorOnboardingForm({ className }: { className?: string }) {
   const [error, setError] = useState<string | null>(null);
 
   const [formData, setFormData] = useState({
+    cf_mentor_name: '',
     cf_mentor_title: '',
     cf_mentor_org: '',
     cf_mentor_bio: '',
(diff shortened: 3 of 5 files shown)


bofengb:
Thanks for getting MD-02 up; the component itself is clean and easy to read. I wanted to flag a few things before approval, because I've been looking at this one alongside the rest of the mentor epic (#90 merged, plus #108, #106, and #111 in flight), and a couple of the findings really only make sense in that wider context.

---

## 1. The Goal

Per issue #63, MD-02 is the user-facing step where a logged-in participant submits their professional details to become a mentor. Each question should map 1:1 to the LearnWorlds Custom User Fields defined in MD-01 (e.g. "Job Title" → `cf_mentor_title`), "Areas of Expertise" should be a multi-select, the page should be reachable to any logged-in user, and submitting the form should update the user's profile data in the backend.

---

### 2.1 The form's submissions never reach `mentor_profiles`

This is the one I'd like to sort before merge. The form calls:

```ts
await supabase.auth.updateUser({
  data: {
    ...formData,
    cf_mentor_expertise: expertise,
    is_mentor_applicant: true,
  },
});
```

That writes to `auth.users.raw_user_meta_data`. But the source of truth for mentors in this repo is the `mentor_profiles` table added in merged PR #90, and I can't find anything in `app/`, `components/`, or `lib/` that reads `cf_mentor_*` or `is_mentor_applicant` back out of auth metadata. Meanwhile:

- The public mentors API in #109 (`app/api/mentors/route.ts`) only reads `mentor_profiles` with `isVisible = true`.
- The admin approval API/UI in #160 / #161 is being built against `mentor_profiles` too.
- There's no code that syncs auth metadata → `mentor_profiles`.

So today a submitted application looks successful to the user but is invisible to every downstream consumer.

**Fix:** route submissions through a server-side API that upserts `mentor_profiles` via Drizzle, using the `LW_FIELD_TO_DB_COLUMN` map from #108. Something along the lines of:

```ts
// app/api/participant/mentor-application/route.ts
import { NextResponse } from 'next/server';
import { authServer } from '@/lib/auth';
import { db } from '@/lib/db';
import { mentorProfiles } from '@/lib/db/schema';

export async function POST(req: Request) {
  const user = await authServer.requireParticipant();
  const body = await req.json();

  await db
    .insert(mentorProfiles)
    .values({
      learnworldsUserId: user.id,
      fullName: body.fullName,
      title: body.cf_mentor_title,
      organization: body.cf_mentor_org,
      bio: body.cf_mentor_bio,
      linkedinUrl: body.cf_mentor_linkedin,
      calendlyUrl: body.cf_mentor_calendly,
      tags: body.cf_mentor_expertise ?? [],
      isVisible: false,
    })
    .onConflictDoUpdate({
      target: mentorProfiles.learnworldsUserId,
      set: {
        /* same fields, omit isVisible so admin approval controls visibility */
      },
    });

  return NextResponse.json({ ok: true });
}
```

Then the form posts to that route instead of `supabase.auth.updateUser`. That also keeps `isVisible = false` until the admin approval flow flips it on, which matches what MD-11/MD-12 seem to expect.

### 2.2 Overlap with the MD-06 webhook data path (#106)

Worth pausing on before this merges. Reading #106 and #108 together, the intended shape of the epic looks like:

1. MD-01 / #108 — defines the LearnWorlds custom fields (`cf_mentor_*`) and the map to `mentor_profiles` columns.
2. MD-06 / #106 — receives the LearnWorlds webhook and upserts `mentor_profiles`.
3. MD-08 / #111 — configures the LearnWorlds side.

Issue #63 itself uses LearnWorlds terminology ("**Form Learning Activity**", "Map each question's answer to the specific Custom User Field"), which reads much more like a LearnWorlds configuration task (sibling to MD-01 and MD-08) than a Next.js form. If we ship this PR as-is, we end up with two parallel entry points — a judge-portal form and a LearnWorlds activity, writing to two different stores, and only the LearnWorlds path actually reaches `mentor_profiles`.

**Fix:** I'd suggest syncing with whoever owns the epic to confirm whether MD-02 is meant to live in judge-portal or in LearnWorlds. If in judge-portal, 2.1 fixes it. If in LearnWorlds, this PR can be closed in favor of configuring the Form Learning Activity there, and MD-06 becomes the single write path.

### 2.3 Field set drifts from MD-01 (#108)

Comparing against `LEARNWORLDS_MENTOR_FIELDS` in #108's `lib/webhooks/learnworlds-fields.ts`:

| Field | In MD-01? | In this PR? | Notes |
|-------|-----------|-------------|-------|
| `cf_mentor_title` | yes | yes | ✓ |
| `cf_mentor_org` | yes | yes | ✓ |
| `cf_mentor_bio` | yes | yes | ✓ |
| `cf_mentor_linkedin` | yes | yes | ✓ |
| `cf_mentor_calendly` | yes | yes | ✓ |
| `cf_mentor_email` | yes (`dbColumn: null`) | yes | ✓ (note: email isn't persisted to `mentor_profiles` by MD-01's map) |
| `cf_mentor_photo` | **yes** | **missing** | maps to `mentor_profiles.photoUrl` — worth adding a URL input |
| `cf_mentor_expertise` | **no** | **yes** | closest column is `mentor_profiles.tags text[]`; suggest renaming to `cf_mentor_tags` or adding to MD-01 |
| `is_mentor_applicant` | no | yes | no consumer; if we need an application-status flag, I'd prefer a dedicated column (e.g. `mentor_profiles.status`) rather than auth metadata |

**Fix:** align the form with `LEARNWORLDS_MENTOR_FIELDS` — add `cf_mentor_photo`, and either drop `cf_mentor_expertise` / `is_mentor_applicant` or extend MD-01 in the same PR so the contract stays in sync.

### 2.4 Important: the page isn't reachable from the participant UI

I couldn't find any link to `/participant/mentor-onboarding` in `app/participant/components/participant-sidebar.tsx` or `app/participant/page.tsx`, so unless a user knows the URL, they'd never get here.

**Fix:** add a "Become a Mentor" entry to the participant sidebar (or a dismissible card on the dashboard) that routes to `/participant/mentor-onboarding`. Ideally hide it for users who already have a `mentor_profiles` row, but even an unconditional link is fine for a first pass.

---

Because MD-02's meaning depends on the LearnWorlds contract, I think the safe merge order is:

1. **#108** (MD-01 — field definitions) — the contract the form has to respect.
2. **#106** (MD-06 — webhook receiver) — defines the canonical write path to `mentor_profiles`.
3. **#111** (MD-08 — LearnWorlds webhook config) — wires LearnWorlds → judge-portal.
4. **#107** (this PR) — only after 1–3, so the form either disappears in favor of a LearnWorlds activity or is wired through an API that actually persists to `mentor_profiles`.
5. **#109** (MD-07 public mentors API), **#95** (MD-09 widget), **#160** / **#161** (MD-11/12 admin) — downstream readers, merge after the write path is settled.

components/mentor-onboarding-form.tsx
```diff
@@ -0,0 +1,223 @@
...
+import { useState } from 'react';
+import { useRouter } from 'next/navigation';
+import { cn } from '@/lib/utils';
+import { createClient } from '@/lib/supabase/client';
+import { Button } from '@/components/ui/button';
+import { Card, CardContent, CardDescription, CardHeader, CardTitle } from '@/components/ui/card';
+import { Input } from '@/components/ui/input';
+import { Label } from '@/components/ui/label';
+import { Textarea } from '@/components/ui/textarea';
+import { Checkbox } from '@/components/ui/checkbox';
```
nsoliven, line 11: I think this is still the main blocker for MD-02. Issue #63 says “Submitting the form updates the user's profile data in the backend,” but this route currently only logs the submitted body and returns `{ ok: true }`.

That means the UI can show “Submission Successful” even though no mentor profile, user metadata, or backend record was actually updated. Can this route persist the submitted application to the intended backend source before returning success, or clearly mark this PR as mock-only if persistence is intentionally out of scope?

nsoliven (commented):
Thanks for updating this after the earlier feedback. The page/form structure itself looks straightforward, and I can see the form now posts to an API route instead of only staying client-side.

I left one main comment because I think the biggest remaining issue is whether the submission actually gets saved anywhere. Once the backend persistence path is real, this should be good.

components/mentor-onboarding-form.tsx
Connor-Palmira: Resolved: Updated API to update the mentor profiles table

sserurich:
> ### 2.2 Overlap with the MD-06 webhook data path (#106)
> Worth pausing on before this merges. Reading #106 and #108 together, the intended shape of the epic looks like:
> 
> 1. MD-01 / [feat(MD-01): LearnWorlds custom field definitions for mentor profiles #108](https://github.com/Freddie-Pike/judge-portal/pull/108) — defines the LearnWorlds custom fields (`cf_mentor_*`) and the map to `mentor_profiles` columns.
> 2. MD-06 / [feat(MD-06): webhook receiver for mentor registration #106](https://github.com/Freddie-Pike/judge-portal/pull/106) — receives the LearnWorlds webhook and upserts `mentor_profiles`.
> 3. MD-08 / [feat(MD-08): Configure LearnWorlds Webhook #111](https://github.com/Freddie-Pike/judge-portal/pull/111) — configures the LearnWorlds side.
> 
> Issue #63 itself uses LearnWorlds terminology ("**Form Learning Activity**", "Map each question's answer to the specific Custom User Field"), which reads much more like a LearnWorlds configuration task (sibling to MD-01 and MD-08) than a Next.js form. If we ship this PR as-is, we end up with two parallel entry points — a judge-portal form and a LearnWorlds activity, writing to two different stores, and only the LearnWorlds path actually reaches `mentor_profiles`.
> 
> **Fix:** I'd suggest syncing with whoever owns the epic to confirm whether MD-02 is meant to live in judge-portal or in LearnWorlds. If in judge-portal, 2.1 fixes it. If in LearnWorlds, this PR can be closed in favor of configuring the Form Learning Activity there, and MD-06 becomes the single write path.

@Connor-Palmira , thanks for updates, however **this point raised 2.2 Overlap with MD-06 webhook path (#106) is Not addressed / still an open design question.** - (This PR implements a judge-portal-native submission path into mentor_profiles. It does not clarify whether MD-02 is supposed to be done in LearnWorlds vs judge-portal, nor does it reconcile “two entry points” (webhook + portal form)). MD-06, has been merged. but will wait to merge this basing on this item. Feel free to clarify.  **Also this second point Page isn’t reachable from participant UI (no navigation link) is not
Not addressed** There is no sidebar/dashboard link change in this PR, so discoverability/navigation is still missing.

nsoliven:
Small concern: this sets `learnworldsUserId` to `user.id`, but `user.id` is the Supabase auth UUID, not necessarily the LearnWorlds user id.

Could this create duplicate mentor profiles if the same user later comes through the LearnWorlds webhook with a real LearnWorlds id? Should we use a separate local user id field or confirm this is intentional?

app/api/mentor-application/route.ts
```diff
@@ -0,0 +1,44 @@
+import { NextResponse } from 'next/server';
+import { db } from '@/lib/db';
+import { mentorProfiles } from '@/lib/db/schema';
+
+export async function POST(req: Request) {
+  try {
+    const body = await req.json();
+
+    const values = {
+      learnworldsUserId: null, // Needs updating when learnworlds user id is fetchable
```
Freddie-Pike: Create Github Issue to update this when Learnworld's connection is fixed.

app/participant/mentor-onboarding/page.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const {
+    data: { user },
+  } = await supabase.auth.getUser();
+
+  if (!user) {
+    redirect('/auth/login');
+  }
+
+  return (
+    <div className="container flex items-center justify-center min-h-[calc(100vh-64px)] py-10">
```
Freddie-Pike, line 16: Check if `min-h-[calc(100vh-64px)]` applies styling.

components/mentor-onboarding-form.tsx
```diff
@@ -0,0 +1,234 @@
...
+
+export function MentorOnboardingForm({ className }: { className?: string }) {
+  const router = useRouter();
+  const [isLoading, setIsLoading] = useState(false);
+  const [showSuccess, setShowSuccess] = useState(false);
+  const [error, setError] = useState<string | null>(null);
+
+  const [formData, setFormData] = useState({
+    cf_mentor_name: '',
```
Freddie-Pike, line 20: Potentially remove `cf` prefix as it's bit hard to know what it means for context?

Freddie-Pike (changes requested):
Minor PR thoughts/comments!

components/mentor-onboarding-form.tsx
tylerbadcock: cf prefix refers to LearnWorlds custom field as described in PR #108. Can add comment for context or look into completely removing the prefix all together.

app/participant/mentor-onboarding/page.tsx
tylerbadcock: Styling appears to be applied

app/api/mentor-application/route.ts
tylerbadcock: Passing learnworldsUserId as null causes a TypeScript error and fails the TypeScript test. With the plan to move away from LearnWorlds, maybe we can look into removing this entirely?

Freddie-Pike: Yeah, let's remove it!

tylerbadcock: I was able to resolve the TypeScript issue. It seems keeping learnworldsUserId is the best option for now as other PRs like MD-06 rely on it, but it can be removed when moving from learnworlds.

Freddie-Pike (approved):

Issue #69, opened 2026-02-17
MD-08 config: configure learnworlds webhook


SaifAhmed1902:
Connect LearnWorlds to the Judge Portal API using the "Tagging Proxy" method, as native Form webhooks don't exist.

### Location
LearnWorlds Admin -> Settings -> Integrations -> Webhooks

### Tasks
- [ ] **Step 1 (The Trigger):** Go to the "Mentor Application" Form activity settings.
  - [ ] Set "Tag user on submission" to `role_mentor`.
- [ ] **Step 2 (The Webhook):** Go to Integrations -> Webhooks.
  - [ ] Create new Webhook.
  - [ ] **Event:** Select `User Tag Added`.
  - [ ] **Target URL:** `https://judgeportal.com/api/webhooks/learnworlds/mentor`.
- [x] **Step 3 (The Filter):** Update the API endpoint code (`MD-06`) to ignore webhooks that DO NOT contain the `role_mentor` tag.

### Acceptance Criteria
- [ ] Submitting the form adds the tag `role_mentor` to the user.
- [ ] The webhook fires immediately after submission.
- [ ] The Judge Portal only processes the request if the `role_mentor` tag is present.

PR #109, opened 2026-03-30
feat(MD-07): Implement Public Mentors API Endpoint

Thomas12556:
Closes #68 

Implemented a public GET endpoint at /api/mentors to support the external LearnWorlds mentor widget in fetching the mentor list.

**Dependencies**
---
 #90 MD-05
- MD-07 is implemented based on the mentorsProfile table created in MD-05.

**Added and Modified Files**
---
- `app/api/mentors/route.ts` (Added)
  - Handles GET requests to fetch mentor data.
  - Filters profiles to only return those where is_visible is true.
- `lib/api/cors.ts` (Added)
  - Contains getCorsHeaders.
  - Future endpoints can use this to maintain consistency. 
- `middleware.ts` (Modified)
  - Updated the matcher/logic to allow public access to /api/mentors by adding the "api/mentors" path.

PR #98, opened 2026-03-27
IC-11 feat: implement LearnWorlds ingestion client  #80 #81

sserurich:
### Title
IC-11 feat: implement LearnWorlds ingestion client and admin trigger endpoint

---

## Summary
This PR implements IC-11 by adding a secure, read-only LearnWorlds ingestion client and an internal admin-triggered ingestion endpoint, with Playwright coverage for endpoint auth and validation behavior.

---

## Related Issue
Closes IC-11 
Closes IC-12 

---

### LearnWorlds Client Service
- OAuth2 client-credentials token flow  
- HTTPS-only configuration validation  
- Required request headers support (`Authorization`, `Lw-Client`)  
- Timeout handling  
- Retry/backoff behavior for transient failures and rate limiting  

### Ingestion Orchestrator
- Creates sync run records  
- Fetches LearnWorlds progress/activity data  
- Stores raw payload records for traceability  
- Updates sync run status and safe error messages  

### Internal Admin Endpoint
- `POST /api/admin/learnworlds/ingest`  
- Admin-only access check  
- Trigger mode validation (`manual`, `scheduled`, `webhook`)  
- Controlled error responses for config/auth/fetch failures  

### Playwright Tests
- Unauthenticated flow (redirect to login via middleware)  
- Authenticated non-admin returns unauthorized  
- Admin invalid trigger mode returns bad request  
- Admin manual trigger returns controlled success/failure response shape  

### Test Utility Refactor
- Extracted user provisioning into `create-test-user` utility  

---

## Why This Change
This establishes a reliable, secure IC-11 ingestion foundation that reads from LearnWorlds only and persists ingestion metadata for downstream processing (IC-13+), while keeping authentication and operational failures controlled and observable.

---

## Validation
- New Playwright spec passes locally after environment setup  
- Lint/type checks for new files pass  
- Endpoint behavior aligns with current middleware redirect model for unauthenticated requests  

---

## Notes
- API remains read-only against LearnWorlds  
- Current unauthenticated API expectation is redirect (middleware behavior), not direct JSON 401  
- Uses existing sync and raw staging tables introduced in prior schema phase  

---

### Functionality
- [ ] Ingestion client authenticates and fetches data correctly  
- [ ] Admin endpoint triggers ingestion successfully  
- [ ] Trigger mode validation works as expected  

### Security
- [ ] Admin-only access enforced correctly  
- [ ] No secrets exposed in logs  
- [ ] HTTPS enforced for outbound calls  

### Reliability
[...]

PR #97, opened 2026-03-27
Feature/learnworlds data ingestion schema  issue #96

sserurich:
## Summary
This PR introduces the LearnWorlds data-ingestion database foundation for the schema-focused issue.

---

## Related Issue
Closes[ IC-19 (LearnWorlds ingestion schema + Drizzle migration setup)](https://github.com/Freddie-Pike/judge-portal/issues/96)

---

## What’s Included
- Added ingestion run tracking table: `learnworlds_sync_runs`  
- Added raw payload staging table: `learnworlds_raw_payloads`  
- Added normalized persistence table: `learner_progress`  
- Added constraints for data integrity (progress bounds, non-negative counters, valid status ranges)  
- Added idempotency constraints for safe re-processing  
- Added indexes for:
  - learner lookup  
  - sync observability  
  - payload access patterns  
- Added feature migration file with the final numbering convention (`005_...`)  

---

## Scope Notes
- This PR is **schema/migration only**  
- No LearnWorlds API client logic is included  
- No transformation service logic is included  
- No recommendation engine logic is included  

---

## Changed Files
- `schema.ts`  
- `005_learnworlds-ingestion-foundation.sql`  

---

## Validation
- Prettier check passed for updated files  
- Schema compiles without TypeScript errors  
- Migration is aligned with existing feature migration structure  

---

## How to Apply Locally
```bash
npm run db:update learnworlds-ingestion-foundation
```
---

### Schema & Migration
- [ ] Tables created correctly (`learnworlds_sync_runs`, `learnworlds_raw_payloads`, `learner_progress`)
- [ ] Constraints enforce valid data (progress bounds, non-negative values, status ranges)
- [ ] Idempotency/uniqueness constraints prevent duplicates
- [ ] Migration runs successfully on clean and existing DB

### Code Quality
- [ ] Drizzle schema matches SQL migration
- [ ] No TypeScript errors
- [ ] Code follows project conventions

### Validation & Testing
- [ ] Tables verified in database after migration
- [ ] Insert/select queries work as expected
- [ ] Invalid data is correctly rejected by constraints

### CI / Formatting
- [ ] Prettier passes
- [ ] Linting passes
- [ ] GitHub Actions checks are green

PR #111, opened 2026-03-30
feat(MD-08): Configure LearnWorlds Webhook

Thomas12556:
Updated the LearnWorlds mentor webhook endpoint so webhooks that do not contain the role_mentor tag are ignored. 

Files Changed
---
- app/api/webhooks/learnworlds/mentor/route.ts
  - If the role_mentor tag is missing, the request is ignored and a 200 OK status is sent back.
- tests/mentor-webhook.spec.ts
  - Updated the validPayload to include the role_mentor tag.
  - Added a new test case: ignores webhook missing the role_mentor tag.

Additional Notes
---
This depends on #106 MD-06 (webhook receiver for mentor registration) as it updates the endpoint to ignore webhooks without the role_mentor tag.

Implements part of https://github.com/Freddie-Pike/judge-portal/issues/69

### DIFF
--- a/app/api/webhooks/learnworlds/mentor/route.ts
+++ b/app/api/webhooks/learnworlds/mentor/route.ts
@@ -43,6 +43,13 @@ export async function POST(request: NextRequest) {
       return NextResponse.json({ error: 'Malformed JSON' }, { status: 400 });
     }
 
+    const tags = payload.tags || [];
+
+    if (!Array.isArray(tags) || !tags.includes('role_mentor')) {
+      console.log('Webhook ignored: User does not have the role_mentor tag.');
+      return NextResponse.json({ message: 'Ignored: No mentor tag' }, { status: 200 });
+    }
+
     const userId = payload.user_id || payload.id;
     const fullName = payload.name || payload.full_name || 'Unknown';
 
--- a/tests/mentor-webhook.spec.ts
+++ b/tests/mentor-webhook.spec.ts
@@ -18,6 +18,7 @@ const validPayload = {
   cf_mentor_bio: '10 years of experience in residential real estate.',
   cf_mentor_linkedin: 'https://linkedin.com/in/janedoe',
   cf_mentor_calendly: 'https://calendly.com/janedoe',
+  tags: ['role_mentor'],
 };
 
 test.describe('Mentor webhook endpoint', () => {
@@ -52,6 +53,18 @@ test.describe('Mentor webhook endpoint with configured secret', () => {
     });
 
     expect(response.status()).toBe(401);
+  });
+
+  test('ignores webhook missing the role_mentor tag', async ({ request }) => {
+    const payloadWithoutTag = { ...validPayload, tags: ['student_tag'] };
+    const response = await request.post(WEBHOOK_URL, {
+      headers: { 'x-lw-signature': signPayload(payloadWithoutTag) },
+      data: payloadWithoutTag,
+    });
+
+    expect(response.status()).toBe(200);
+    const body = await response.json();
+    expect(body.message).toBe('Ignored: No mentor tag');
   });
 
   test('rejects payload missing user_id', async ({ request }) => {


bofengb:
# PR Review: `feat/MD-08-configure-learnworlds-webhook`

Hi, thanks for getting this one out the door. I reviewed this alongside the MD-05 PR (#90, already merged), the MD-06 webhook receiver (#106, still open), and the MD-07 PR (#109, also still open) so I could get the full picture of how these mentor PRs fit together. Most of my comments below are about the *shape* of this PR rather than the MD-08 logic itself. The actual change you're trying to make is tiny and correct, but it's currently hidden inside ~270 lines of work that belongs to other branches.

---

## 1. The Goal

Issue #69 (MD-08) asks us to configure the LearnWorlds webhook so that only users with the `role_mentor` tag are written into `mentor_profiles`. The PR description matches that perfectly: "webhooks that do not contain the `role_mentor` tag are ignored." In the ideal final diff, this PR should be around ten lines of code:

- one tag-check block inside `app/api/webhooks/learnworlds/mentor/route.ts`
- one `tags: ['role_mentor']` addition to the shared test payload
- one new test case

Everything else should come in via the upstream PRs this depends on.

---

### 2.1 The PR contains MD-05 + MD-06 + MD-08 all at once

When I pulled the diff, I saw six files changed, but only two of them are actually MD-08:

- `app/api/webhooks/learnworlds/mentor/route.ts` (the `role_mentor` check block)
- `tests/mentor-webhook.spec.ts` (the new "ignores webhook missing the role_mentor tag" test)

The remaining changes belong to already-merged or still-open work:

| File | Belongs to |
| --- | --- |
| `supabase/migrations/features/004_mentor_profiles.sql` | MD-05 / PR #90 (already merged) |
| `lib/db/schema.ts` (the `mentorProfiles` block) | MD-05 / PR #90 (already merged) |
| `app/api/webhooks/learnworlds/mentor/route.ts` (the file itself, minus the tag check) | MD-06 / PR #106 (still open) |
| `app/api/admin/metrics/route.ts` (deletion) | MD-05's internal review history — unrelated to MD-08 |
| `package-lock.json` | unrelated noise (see §2.6) |

I think what happened is that the branch was cut before #90 merged and has drifted. The fix is a rebase.

**Fix:** after rebasing, the PR diff should collapse to the ~10 MD-08-specific lines. If the diff is still large after rebasing, something went wrong in the rebase.

---

### 2.2 `004_mentor_profiles.sql` already exists on `main`

Our project conventions require a new sequential prefix for each feature migration. Main currently has `001`..`005`, and `004_mentor_profiles.sql` is already taken by the MD-05 migration that merged on 2026-04-15. This PR proposes a second, *different* file at the same path, which will either fail on duplicate objects or silently diverge when someone runs `npm run db:setup` after a merge.

A few concrete differences between the two `004` files worth calling out:

| | `main` | This PR |
| --- | --- | --- |
| `is_visible` default | `FALSE` | `TRUE` |
| `DROP TRIGGER IF EXISTS` guard before `CREATE TRIGGER` | present (idempotent) | missing |

**Fix:** drop the migration change entirely after the rebase. `004` is taken and the schema it installs is already live.

---

### 2.3 `is_visible` default flipped from `FALSE` to `TRUE` (security-relevant)

This one is the most concerning because it's subtle. Main's schema has:

```ts
isVisible: boolean('is_visible').default(false).notNull(),
```

and the corresponding migration has `is_visible BOOLEAN DEFAULT FALSE NOT NULL`. That `FALSE` default is the entire mechanism behind the "pending admin approval" flow, because the RLS policy `"Public can view visible mentor profiles"` exposes any row where `is_visible = true`, a `FALSE` default guarantees new mentors are invisible until an admin flips the bit.

This PR changes both sides to `TRUE`:

```ts
// in this PR
isVisible: boolean('is_visible').default(true).notNull(),
```

```sql
-- in this PR
is_visible BOOLEAN DEFAULT TRUE NOT NULL,
```

The webhook handler itself passes `isVisible: false` on insert, so the regression is *masked* when the webhook is the only entry point, but any other insertion path (seed data, admin UI, future migrations) would silently make mentors public. I'd strongly prefer we never reintroduce this change.

**Fix:** after the rebase this falls out naturally — the schema and migration changes disappear, and `main`'s `FALSE` default stays in place. No independent action needed.

---

### 2.4 Byte-for-byte duplication with the still-open #106

The `app/api/webhooks/learnworlds/mentor/route.ts` file in this PR is the same file that PR #106 adds, plus your 5-line tag-check block. Five of the six tests in `tests/mentor-webhook.spec.ts` are also from #106. If #106 merges first, your diff collapses to a clean, easy-to-review delta. If this PR tries to land first, it silently pre-empts #106 and you inherit responsibility for MD-06.

**Fix:** base this branch on `feat/MD-06-mentor-webhook` rather than `main` so reviewers see only the MD-08 delta:

```bash
gh pr edit 111 --base feat/MD-06-mentor-webhook
```

Once #106 merges to `main`, rebase back onto `main` and the diff stays tiny.

---

### 2.5 `mergeable: CONFLICTING`

GitHub reports the PR as conflicting, which is the natural consequence of §2.1–§2.3 — the branch is re-adding tables and files that `main` already has. Concretely, `main` has moved on since this branch was cut: #90 (MD-05), #97 (LearnWorlds ingestion schema), and #98 (LearnWorlds ingestion client) have all merged.

**Fix:** the cleanest recovery I can think of is:

```bash
git fetch origin
git checkout feat/MD-08-configure-learnworlds-webhook
git rebase origin/main
# during the rebase, drop the schema / migration / route / admin-metrics / package-lock.json hunks
# keep only the tag-check block and the new test case
```

After that, consider the base-branch change from §2.4 so #106 lands first.

---

1. **#106 (MD-06 webhook receiver)** — merge first. It has its own `CONFLICTING` status to resolve, but the webhook file genuinely belongs here.
2. **#111 (this PR, MD-08)** — rebase onto `main` after #106 lands; diff collapses to ~10 lines.
3. **#109 (MD-07 public mentors API)** — same author, same scope-creep pattern (duplicate `004_mentor_profiles.sql`, duplicate `admin/metrics` deletion, `package-lock.json` noise). The same rebase guidance applies there.

sserurich (approved):
Thanks @Thomas12556 , the review feedback issues have been addressed.

PR #151, opened 2026-04-14
feat(seed): add LearnWorlds ingestion, transformation, and recommenda…

sserurich:
## PR Title
feat(seed): add end-to-end test seed data for LearnWorlds ingestion, transform, and personalized recommendations

---

## PR Description
Closes #129 

---

### Summary
This PR expands `seed.ts` to populate realistic test data across the newly added LearnWorlds and personalized recommendation schemas, in addition to existing event/team/criteria seed data.

---

### What Changed

**Environment Setup**
- Added robust env loading for seed execution:
  - Loads `.env.local` and `.env` in the seed script
  - Supports standard DB URL env keys and validates presence

**RLS Policy Handling**
- Made RLS policy setup idempotent:
  - Drops existing policies before re-creating them
  - Allows repeated seed runs without policy conflicts

**LearnWorlds Ingestion & Transformation Seeding**
- Seeded:
  - `learnworlds_sync_runs`
  - `learnworlds_raw_payloads` (valid + invalid examples)
  - `learner_progress` (with upsert behavior)

**Personalized Recommendation Data Seeding**
- Seeded:
  - `learning_items`
  - `learner_recommendations`
  - `recommendation_impressions`
  - `recommendation_outcomes`
  - `recommendation_feedback`
  - `learner_item_events`
  - `ml_training_examples`
  - `model_registry`

**Seed Execution Improvements**
- Unified inserts through one SQL client path in seed script:
  - Avoids mixed connection behavior
  - Keeps seed flow consistent and deterministic

---

### Why

To provide reusable, realistic test fixtures for:

- LearnWorlds ingestion pipeline  
- Transformed learner progress  
- Recommendation generation  
- Impression/outcome tracking  
- ML training dataset generation  

This enables faster and more reliable local/dev and test DB setup for API, analytics, and ML experimentation.

---

### Issue Coverage

This PR provides seeded data support for:

 - **IC-19** → learnworlds ingestion foundation schema  
 - **IC-26** → learner_recommendations  
- **IC-30** → learner_item_events  
- **IC-31** → recommendation_impressions, recommendation_outcomes  
- **IC-32** → recommendation_feedback  
- **IC-34** → learning_items, ml_training_examples  
- **IC-45** → model_registry  

---

### Validation
- `npm run lint` passed  
- `npx tsx lib/db/seed.ts` executed successfully  
- Seed logs confirm all datasets created end-to-end  
---

### Notes

- Seed values include a timestamp-based suffix to reduce collision risk across repeated runs  
[...]

PR #154, opened 2026-04-15
IC-27 test (seed): create simulated dataset for LearnWorlds ingestion and recommendation scenarios

zsin520:
Test (seed): generates synthetic data to test recommendation model

## Description
Generates a robust set of fake users to test the different branches of the recommendation model.

## Summary
Updated the code from PR #151  to generate a robust, 4-tier matrix of learner states. This expands the test data beyond simple sparse progress to properly test the diverse routing branches (Time, Frequency, and Density). The new code tracks several different dimensions:
1. Type of users
2. Level of activity
3. Course selection

## Rationale
The previous seed data would fail to evaluate the recommendation model at every branch. A hybrid architecture that combines a rule-based and machine learning recommendation system will have multiple branches. Multiple user types will give us an accurate picture of how the model performs for each type of learner. 

## Notes
Course names were generated with Gemini: https://gemini.google.com/share/b442d9abdf7b. Only a subset of courses created were used to simplify testing.

### DIFF
--- a/lib/db/recommendation-model-seed.ts
+++ b/lib/db/recommendation-model-seed.ts
@@ -16,15 +16,15 @@ if (!connectionString) {
 }
 
 const seedDatabase = async () => {
-  console.log('Setting up RLS policies and seeding database...');
+  console.log('Seeding recommendation model test data...');
   const seedTag = Date.now().toString();
 
   /*
     Define four different user types
-        1. Cold start: rule-based (default)
-        2. Dropout: started course, but hasn't returned in a long time, rule-based
-        3. Casual: started course, completed it, but data is sparse, rule-based
-        4. Power: frequent user, has completed courses, ML
+        1. Cold start: no history → default_popular fallback rule
+        2. Dropout: started course, hasn't returned in 60 days → resume_inactivity rule
+        3. Casual: completed one course, sparse data → next_step_progression rule
+        4. Power: dense recent history, 5+ events → ML path (falls back to high_progress in Phase 1)
   */
   const userColdStart = `lw-user-${seedTag}-cold`;
   const userDropout = `lw-user-${seedTag}-dropout`;
@@ -32,7 +32,9 @@ const seedDatabase = async () => {
   const userPower = `lw-user-${seedTag}-power`;
 
   /*
-    Define Course IDs corresponding to the catalog
+    Define Course IDs matching the catalog below.
+    All tagged with seedTag to avoid collisions across runs.
+    'BIO-340' is a static item (no tag) that matches DEFAULT_FALLBACK_ITEM_ID in RULES_CONFIG.
   */
   const courseEpi = `BIO-340-${seedTag}`;
   const course3D = `BIO-450-${seedTag}`;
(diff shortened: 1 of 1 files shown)


lolbro2k (changes requested):
Thank you for this PR zach,

I think it would be better to add information or indicators that this data is seed data. If used in actual implementation this data and real-world data would be indistinguishable without reading this file. Adding small information like "seed" to parts of this data would be useful for future use. 

### Changes Requested
Add some indicator that this weed data is simulated and not real-world data because in a supabase table it would all look too similar.

sserurich (commented):
Hi @zsin520, this works well — it successfully seeds the data. Nice work  
There are a few small adjustments that could improve safety and robustness.

---

### ✅ What’s Good

- **4-tier user matrix is well-designed**  
  Cold Start, Dropout, Casual, and Power users cover all routing branches of a hybrid rule-based + ML system as described.

- **`seedTag` timestamp isolation**  
  User IDs and course IDs are namespaced with `Date.now()`, ensuring repeated runs don’t collide. This is a solid pattern for test data.

- **Temporal events are realistic**  
  - Dropout user stuck ~60 days ago  
  - Power user activity dense within last 7 days  
  This correctly exercises **Time / Frequency / Density** signals.

- **ML pipeline is fully exercised**  
  Training examples and model registry entries are populated — not just recommendation outputs. This is thorough.

- **Fallback → Rule → ML progression**  
  The 4 user types map directly to the intended model architecture.

---

### ⚠️ Issues / Feedback

#### 1. RLS policy management is risky

The script runs `DROP POLICY IF EXISTS` + `CREATE POLICY` on the table during every seed run.

- This modifies shared RLS policies
- If run against a shared DB, it could temporarily remove access control

👉 These policies should live in **migrations**, not seed scripts.

---

#### 2. No `ON CONFLICT` / cleanup guard

The following inserts do not include conflict handling:

- `learning_items`
- `learner_progress`
- `learner_item_events`

While `seedTag` reduces collision risk, inserts could still fail if:
- The script runs twice in quick succession
- Timestamps collide at the millisecond level

---

#### 3. No `package.json` script entry

There is no script like:

```json
"db:seed:recommendation": "tsx lib/db/recommendation-model-seed.ts"
```


### **How to Fix ***
#### Fix 1 — Remove RLS policy management from seed
Delete the DROP POLICY block entirely. Those belong in a migration (Explore more on this to be certain, just suggestion)

#### Fix 2 — Add ON CONFLICT DO NOTHING to all inserts
Append ON CONFLICT DO NOTHING to every raw SQL insert:
```
await client`
  INSERT INTO learning_items (item_id, item_type, title, description, category, difficulty_level, is_active)
  VALUES (...)
  ON CONFLICT (item_id) DO NOTHING
`;

await client`
  INSERT INTO learner_progress (learnworlds_user_id, course_id, progress_percentage, completion_status)
  VALUES (...)
  ON CONFLICT (learnworlds_user_id, course_id) DO NOTHING
`;

// same for learner_item_events, ml_training_examples, learner_recommendations```
```

#### Fix 3 — Add [package.json script OR Put the block in already defined seed.ts
```"db:seed:recommendation": "tsx lib/db/recommendation-model-seed.ts"
```

zsin520:
This code has been added to PR #156, so this PR will be closed.

PR #150, opened 2026-04-14
feat: add personalized recommendation schema foundation (IC-26,  IC-29, IC-30, IC-31, IC-32, IC-34, IC-45)

sserurich:
## PR Title
feat: add personalized recommendation schema foundation (IC-26,  IC-29, IC-30, IC-31, IC-32, IC-34, IC-45)

---

## PR Description

Closes #128, #131, #132, #133, #134, #136, and #149.

### Summary

Implemented the personalized recommendation and data-science schema foundation for the LMS.

### What’s Included

Added new feature migration for recommendation / data-science tables:
- learner_recommendations
- learner_item_events
- recommendation_impressions
- recommendation_outcomes
- recommendation_feedback
- learning_items
- ml_training_examples
- model_registry

Added matching Drizzle schema definitions and exported types for all of the above.

### Design & Consistency

- Kept naming aligned with LearnWorlds integration using:
  - `learnworlds_user_id` (SQL)
  - `learnworldsUserId` (Drizzle)

- Maintained consistency with existing migration patterns:
  - idempotent table/index creation
  - UTC timestamp defaults
  - structured, sectioned migration layout
  - success logging

### Data Integrity & Constraints

Added constraints and indexes to ensure quality and performance:
- enum-like validation using CHECK constraints (event_type, source, outcome_type, etc.)
- rating bounds (1–5) and non-negative checks (scores, counts, ranks)
- composite indexes for user + time access patterns
- foreign key relationships:
  - recommendation → impressions → outcomes
  - self-referential prerequisite relationship in `learning_items`

### Issue Coverage

This PR provides the schema foundation for:

- **IC-26** → learner_recommendations  
- **IC-30** → learner_item_events  
- **IC-31** → recommendation_impressions, recommendation_outcomes  
- **IC-32** → recommendation_feedback  
- **IC-34** → learning_items, ml_training_examples  
- **IC-45** → model_registry  

### Validation

- Lint passed after schema integration  
- Migration follows existing repo conventions  
- Tables and indexes verified locally

PR #156, opened 2026-04-15
IC-25 feat: implement rule-based recommendation engine and catalog

### LINKED ISSUE #127: IC-25 feat: implement rule-based recommendation engine and catalog
### Description
Build the initial recommendation engine using deterministic rules and a configurable recommendation catalog. Relates to #88 
 
### Scope
- Create recommendation catalog (mapping next items)
- Implement rules:
  - resume after inactivity
  - next-step progression
  - high-progress → advanced item
  - fallback
- Ensure deterministic rule priority

### Acceptance Criteria
- Recommendation catalog exists and is configurable  
- Engine returns exactly one recommendation per learner  
- Rules execute in defined priority order  
- Each recommendation includes:
  - recommended item  
  - rationale  
  - rule_matched

zsin520:
## Description
Closes #127  IC-25. Implements the rule-based part of the recommendation model. The changes define a new function to generate recommendations for users with sparse or no data `generateRuleBasedRecommendation`. 

## Summary
The model is implemented in the file `lib/services/recommendation-model.ts`. It defines the rules used to provide recommendation for users with sparse or no data. Briefly:
1. Loads latest activity for the learner and the user's course progress
2. Based on the user's activity it returns a recommendation object, which is also inserted into the recommendation table for tracking

The model makes a recommendation based on the following rules, where the cold-start user is the default behavior
1. A cold-start user with no activity - recommends default course 
2. A user with a history of activity, but no login within the last X days - recommends the last course the user interacted with
3. A user with recent login activity and a course near completion - recommends completing the course with the most progress
4. A user that recently completed a course - recommends the next course in the series

## Notes
- There are several important threshold that are defined: INACTIVITY_THRESHOLD_DAYS, HIGH_PROGRESS_THRESHOLD, DEFAULT_FALLBACK_ITEM_ID

### DIFF
--- a/.github/workflows/lint-and-format.yml
+++ b/.github/workflows/lint-and-format.yml
@@ -2,7 +2,6 @@ name: Lint & Format
 
 on:
   pull_request:
-  push:
 
 jobs:
   lint-and-format:
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -2,7 +2,6 @@ name: Playwright Tests
 
 on:
   pull_request:
-  push:
 
 jobs:
   playwright-tests:
@@ -63,15 +62,15 @@ jobs:
             kill $(cat .server.pid) || true
           fi
       - uses: actions/upload-artifact@v4
-        if: ${{ !cancelled() }}
+        if: failure()
         with:
           name: playwright-report
           path: playwright-report/
-          retention-days: 30
+          retention-days: 3
       - name: Upload server logs
-        if: ${{ !cancelled() }}
+        if: failure()
         uses: actions/upload-artifact@v4
         with:
           name: server-logs
           path: server.log
-          retention-days: 7
+          retention-days: 3
--- a/app/admin/components/competition-overview.tsx
+++ b/app/admin/components/competition-overview.tsx
@@ -1,0 +1,145 @@
+'use client';
+
+import { useState, useEffect } from 'react';
+import { Card, CardContent, CardHeader, CardTitle } from '@/components/ui/card';
+import { Badge } from '@/components/ui/badge';
+import { Trophy, Loader2 } from 'lucide-react';
+
+interface Competition {
+  id: string;
+  eventId: string;
+  eventName: string;
+  eventStatus: string;
+  title: string | null;
+  prize: string | null;
+  tags: string[] | null;
+  deadline: string | null;
+  country: string | null;
+  challengeType: string;
+  shortDescription: string | null;
+  participantSignupUrl: string | null;
+}
+
+export default function CompetitionOverview() {
+  const [competitions, setCompetitions] = useState<Competition[]>([]);
+  const [isLoading, setIsLoading] = useState(true);
+
+  useEffect(() => {
+    const fetchCompetitions = async () => {
+      try {
+        const response = await fetch('/api/admin/competitions');
+        if (response.ok) {
+          const data = await response.json();
+          setCompetitions(data);
+        }
+      } catch (error) {
+        console.error('Error fetching competitions:', error);
+      } finally {
+        setIsLoading(false);
+      }
+    };
+
+    fetchCompetitions();
+  }, []);
+
+  if (isLoading) {
+    return (
+      <div className="flex items-center justify-center py-20">
+        <Loader2 className="h-6 w-6 animate-spin text-muted-foreground" />
+      </div>
+    );
+  }
+
+  if (competitions.length === 0) {
+    return (
+      <div className="text-center py-20">
+        <Trophy className="h-10 w-10 mx-auto mb-3 text-muted-foreground/40" />
+        <p className="text-muted-foreground">
+          No competitions yet. Promote an event to a competition from the Events tab.
+        </p>
+      </div>
+    );
+  }
+
+  return (
+    <div className="space-y-4">
+      <div>
+        <h2 className="text-xl font-semibold text-foreground">Competitions</h2>
+        <p className="text-sm text-muted-foreground">
+          Overview of all events promoted to competitions.
+        </p>
+      </div>
+
+      <div className="grid gap-4">
+        {competitions.map((competition) => (
+          <Card key={competition.id}>
+            <CardHeader className="pb-3">
+              <div className="flex items-start justify-between gap-4">
+                <div>
+                  <CardTitle className="text-base">
+                    {competition.title || competition.eventName}
+                  </CardTitle>
+                  <p className="text-sm text-muted-foreground mt-0.5">
+                    Event: {competition.eventName}
+                  </p>
+                </div>
+                <Badge variant="outline" className="capitalize shrink-0">
+                  {competition.eventStatus}
+                </Badge>
+              </div>
+            </CardHeader>
+            <CardContent className="space-y-3">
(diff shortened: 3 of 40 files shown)


lolbro2k (changes requested):
Thanks for this PR I like the modifiability.
### Notes
Our architecture is meant for a rules-based and ML model for recommendation system. I think this PR should be adapted to also take in the ML's recommendations and then apply rules for a final submission. 

From my understanding this is just made to work by applying rules if there isn't enough courses completed for a ML model which I think it applied fairly robustly. It is simple and efficient. I, however, think this engine has two features in the architecture
## 1
Apply rules for the recommendation if we cannot use the ML models
## 2 
Apply rules for the recommendation when we can use the ML models
## Changes
the ML model return relevant courses by using a quantized cross-encoder. The rules based engine should take these recommendation and apply rules like, too high difficulty or already completed. LMS systems are usually mostly rules based system with very little ML only for similarity, but the ultimate decision need to be coded by humans.

### end
Overall great PR and I like the robust approach!

sserurich (commented):
Thanks @zsin520 , this is  solid implementatin, however a few things need to be addressed.

### ✅ What’s Good

- **Clean rule cascade**  
  The 4 rules fall through correctly via `if (!recommendation && ...)` guards. Each rule is mutually exclusive and executed in the right order.

- **Correct inactivity logic**  
  Checking `isEventItemInProgress` before applying the inactivity rule correctly avoids recommending a completed course.

- **Efficient prerequisite lookup**  
  Using `inArray` avoids N+1 queries and ensures a single efficient lookup for completed courses.

- **Centralized configuration**  
  The `RULES_CONFIG` constant prevents magic numbers from being scattered across the codebase and makes tuning easier.

- **Fail-fast fallback behavior**  
  The fallback throws if `DEFAULT_FALLBACK_ITEM_ID` doesn’t exist in the DB, ensuring failures are explicit rather than silent.

---

### ❌ Issues / Feedback

#### 1. Missing schema definitions

`schema.learnerItemEvents`, `schema.learnerProgress`, and `schema.learningItems` do not exist in `schema.ts`.

- The schema on this branch only goes up to `competitions`
- None of the queried tables are defined in Drizzle
- This will fail at runtime with a property access error

👉 A migration and corresponding schema definitions must be included in this PR. (Merge the recent Main, the PR defining these has been merged into main ([https://github.com/Freddie-Pike/judge-portal/pull/150](https://github.com/Freddie-Pike/judge-portal/pull/150)) 

---

#### 2. Missing import dependency

`RecommendationResult` is imported from `./recommendation-orchestrator`, which does not exist on this branch.

- `git ls-tree` shows only `rule-based-recommendation-model.ts` under services
- The orchestrator file is missing

👉 Either define the type locally or include the orchestrator file in this PR.

---

#### 3. Rule 1 (inactivity) edge case

`latestEvent.itemId` may refer to a completed course.

- `latestEvent` is the most recent event regardless of status
- If the user completed the course, `isEventItemInProgress` correctly prevents recommendation
- However, if the latest event was a "view" on a course later completed, it still evaluates to false

👉 The logic seems sound, but the comment says *"last course the user was interacting with"*, not *"last in-progress course"* — worth confirming intended behavior.

---

#### 4. Hardcoded fallback ID

```ts
DEFAULT_FALLBACK_ITEM_ID: 'BIO-340'
```
- Couples the model to a specific course ID  
- If the course is removed or renamed, fallback breaks at runtime  

👉 Recommend moving this to:

- Environment variable, or  
- Configurable DB value  

---

#### 5. Missing persistence layer

The PR description states:

> "is also inserted into the recommendation table for tracking"

However:

- The function only returns the recommendation object  
- No insert into `learnerRecommendations` is present  

👉 Either:

- The description is incorrect, or  
- The persistence logic is missing  

Needs clarification.

---

#### 6: No unit tests

- The logic has 4 clear and testable code paths  
- Can be easily tested with mocked DB calls  

👉 Given this is core model logic, unit tests should be included in this PR.

---

### 🧾 Summary

Strong rule-based foundation with clear structure and good design decisions.

Key gaps to address:

- Missing schema + migrations  
- Missing dependency (`recommendation-orchestrator`)  
- Edge-case clarity in Rule 1  
- Hardcoded fallback configuration  
- Missing persistence layer  
- Lack of unit tests  

Once these are resolved, this will be much more robust and production-ready.

zsin520:
Hi @sserurich thank you for the comments,
1. Main branch has been merged
2. Recommendation model files have been consolidated
3. Comment has been added to rule 1 for clarity
4. Variable has been added for hardcoded default
5. Persistence logic is in the orchestrator now, the PR description was outdated

## Testing
A test script has been implemented in `tests/rule-based-recommendation-model.spec.ts`
- all the rules are tested and the error cases
- `vi.mock` from `vitest ` is used so their is no db connection
- run with `npx vitest run tests/rule-based-recommendation-model.spec.ts`

sserurich:
> Hi @sserurich thank you for the comments,
> 
> 1. Main branch has been merged
> 2. Recommendation model files have been consolidated
> 3. Comment has been added to rule 1 for clarity
> 4. Variable has been added for hardcoded default
> 5. Persistence logic is in the orchestrator now, the PR description was outdated
> 
> ## Testing
> A test script has been implemented in `tests/rule-based-recommendation-model.spec.ts`
> 
> * all the rules are tested and the error cases
> * `vi.mock` from `vitest ` is used so their is no db connection
> * run with `npx vitest run tests/rule-based-recommendation-model.spec.ts`

Confirmed: All previously raised PR comments have been addressed.

---

### ✅ Key Fixes Verified

- **Schema definitions**
  - All required tables are now defined in `schema.ts`
  - Supporting migrations are present and correctly implemented

- **Import dependencies**
  - Missing imports between orchestrator and rule-based model resolved

- **Rule 1 (inactivity edge case)**
  - Guard condition implemented
  - Clarifying comment added
  - Edge case covered with test

- **Fallback configuration**
  - Hardcoded ID replaced with environment-driven configuration
  - Default still exists (`BIO-340`) when env var is unset

- **Persistence layer**
  - Recommendation persistence implemented
  - Insert logic added and invoked across both ML and rule-based paths

- **Test coverage**
  - Comprehensive unit tests added
  - Covers core flows and edge/error cases
  - Test run verified: all tests passing (1 file, 6 tests)

---

### 💡 Follow-up (Testing Strategy Alignment)

- Since the project already uses Playwright for end-to-end testing, recommend creating a follow-up issue to:
  - Add Playwright tests for recommendation flows (API + integration level)
  - Potentially align or consolidate testing strategy so Playwright serves as the primary validation layer for end-to-end behavior
  - I have opened up a follow up issue to be addressed in the future #194 

👉 This will ensure consistency with the existing testing approach and provide stronger system-level validation.

---

## 🚀 Final Assessment

- All blocking issues are resolved  
- Implementation is now consistent, test-backed, and production-ready

sserurich (approved):
Looks good to me

sserurich:
> Looks good to me
- CI tests are failing as you communicated to me, let's first address this before we merge

sserurich:
> > Looks good to me
> 
> * CI tests are failing as you communicated to me, let's first address this before we merge

Resolved the small issue that was causing the tests to fail. @zsin520 , going to merge.

PR #159, opened 2026-04-15
Reduce GitHub Actions storage and minute usage

bofengb:
Our Actions storage is running low. This PR applies three small changes to optimize the storage:

- **Remove `push:` trigger from both workflows.** Now they run once per PR update.
- **Upload Playwright artifacts only on failure.** From now on, artifacts are only kept when we actually need them for debugging.
- **Drop artifact retention from 30/7 days to 3 days.** Failed-run artifacts are useful for a day or two, not a month.

### DIFF
--- a/.github/workflows/lint-and-format.yml
+++ b/.github/workflows/lint-and-format.yml
@@ -2,7 +2,6 @@ name: Lint & Format
 
 on:
   pull_request:
-  push:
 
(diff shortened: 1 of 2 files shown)


Freddie-Pike (approved):
Small, big impact changes. Thanks for doing this up bud!

PR #166, opened 2026-04-19
Ic 26 freshness policy

lolbro2k:
### Description
Built on feature/personalized-learning-recommendation-schema to obtain the schema data. to add checks for the recommendation recency and making sure the recommendation stays fresh.

### Summary of Changes
* **Retrieval Logic:** Checks to see if the recommendation is over certain hours old before it attempts to create a new RulesBasedRecommendation to be added into the db. If there is a fresh recommendation available, the logic will use this instead.

### Scope
- [x] Store recommendations in `learner_recommendations`
- [x] Implement freshness rule (e.g. 24 hours)
- [x] Retrieve latest recommendation if valid

### Acceptance Criteria
- [x] Recommendations are persisted in DB  
- [x] Latest recommendation can be retrieved per learner  
- [x] Fresh recommendations are reused  
- [x] Expired recommendations are regenerated 
-


8bit-Apollo (changes requested):
Judge-portal uses \supabase\migrations\consolidated_setup.sql as a comprehensive setup file for all migrations. 
**supabase/migrations/features/006_personalized-learning-recommendation-foundation.sql** needs to be merged into consolidated_setup.sql.

lib/services/recommendation-model.ts
```diff
@@ -22,6 +23,35 @@ export type RecommendationResult = {
   source: 'rule' | 'fallback';
 };
 
+/*
+    Freshnesss Policy:
```
8bit-Apollo: typo - Freshness

lib/services/recommendation-model.ts
```diff
@@ -22,6 +23,35 @@ export type RecommendationResult = {
...
+): Promise<RecommendationResult | null> {
+  const [latestRec] = await db
+    .select()
+    .from(schema.learnerRecommendations)
+    .where(eq(schema.learnerRecommendations.learnworldsUserId, learnworldsUserId))
+    .orderBy(desc(schema.learnerRecommendations.createdAt))
+    .limit(1);
+
+  if (latestRec && latestRec.createdAt) {
+    const generated_at = new Date(latestRec.createdAt).getTime();
```
8bit-Apollo: consider using generatedAt instead of CreatedAt to maintain consistency.

8bit-Apollo (commented):
Other than the issue with supabase/migrations/features/006_personalized-learning-recommendation-foundation.sql mentioned above, this PR looks good. Just fix that and the issues below and this should be ready to merge.

Connor-Palmira (commented):
This PR is a solid implementation of recommendation persistence. There were some small details I noticed that you might want to address.

One thing I noticed is that you mention in your PR that you’re example time period for your freshness policy is 24 hours while you have in `lib/services/recommendation-model.ts`, ` ‘HOURS_FRESHNESS: 72` and I just wanted to make sure that these two were aligned.

Something else that I noticed is that in `learnerRecommendations` in `lib/db/schema.ts` you have an `expiresAt` timestamp but the current logic in your freshness policy calculates freshness manually using `(current - generated_at) / (1000 * 3600) < RULES_CONFIG.HOURS_FRESHNESS)`. You should consider looking into either continuing with manual checking the freshness here and getting rid of the `expiresAt` timestamp or checking to see if the current time exceeds `expiresAt`.

Overall, a really good job.

sserurich (approved):
Thanks @lolbro2k , the feedback left has been addressed and this PR has been scoped now to only the relevant changes.

PR #168, opened 2026-04-19
IC 23/24 - Adds Endpoints for Recommendation Widget

Connor-Palmira:
This PR implements two endpoints for the recommendation widgets to receive recommendation information from Supabase and to submit feedback on recommendations.

Changes
- `GET /api/recommendations`: Endpoint that fetches the most recent recommendation for the user, includes a fallback recommendation for display
- `POST /api/recommendation/feedback`: Endpoint that stores user's feedback information
- Removal of mock data in favor of using data in Supabase instead

Notes
- Testing was done using schema injected from PR #150
- Page can be accessed by adding `/recommendations` to the participant home page
- Testing was done on the following data using Supabase's SQL Editor:
```
CREATE TABLE IF NOT EXISTS "recommendation_feedback" (
  "id" uuid PRIMARY KEY DEFAULT gen_random_uuid(),
  "recommendation_id" uuid REFERENCES "learner_recommendations"("id") ON DELETE CASCADE,
  "learnworlds_user_id" text NOT NULL,
  "recommended_item_id" text NOT NULL,
  "feedback_type" text,
  "rating" integer,
  "comment" text,
  "submitted_at" timestamptz NOT NULL DEFAULT timezone('utc'::text, now()),
  "created_at" timestamptz NOT NULL DEFAULT timezone('utc'::text, now()),

  CONSTRAINT "check_recommendation_feedback_feedback_type" 
    CHECK ("feedback_type" IS NULL OR "feedback_type" IN ('helpful', 'not_helpful')),
  CONSTRAINT "check_recommendation_feedback_rating" 
    CHECK ("rating" IS NULL OR ("rating" >= 1 AND "rating" <= 5))
);

CREATE INDEX IF NOT EXISTS "idx_recommendation_feedback_recommendation" 
  ON "recommendation_feedback" ("recommendation_id");

CREATE INDEX IF NOT EXISTS "idx_recommendation_feedback_learnworlds_user" 
  ON "recommendation_feedback" ("learnworlds_user_id");

ALTER TABLE "recommendation_feedback" ENABLE ROW LEVEL SECURITY;

CREATE POLICY "Users can manage own feedback" ON "recommendation_feedback"
  FOR ALL USING (auth.uid()::text = learnworlds_user_id);

INSERT INTO learner_recommendations (
    learnworlds_user_id,
    recommended_item_id,
    recommended_item_type,
    recommended_title,
    rationale,
    source,
    rule_matched,
    model_version,
    score
)
VALUES 
(
    'ce5cce86-f945-4355-83a6-67171f66d4e6', 
    'li-lesson-456', 
    'lesson', 
    'Advanced Competitive Analysis', 
    'Since you just finished "Market Research Basics," this lesson will help you apply those skills to competitor tracking.', 
    'rule_engine', 
    'sequential_progression', 
    'v1.2.0', 
    0.95
),
(
    'ce5cce86-f945-4355-83a6-67171f66d4e6', 
    'li-course-789', 
    'course', 
    'UI Design for Product Managers', 
    'Your progress in Product Strategy suggests you might benefit from understanding the design hand-off process.', 
    'ml_model', 
    'collaborative_filtering', 
    'v2.0.1-beta', 
    0.82
),
(
    'ce5cce86-f945-4355-83a6-67171f66d4e6', 
    'li-module-101', 
    'module', 
    'Stakeholder Management Workshop', 
    'You are 80% through this module! Only 15 minutes left to earn your completion badge.', 
    'rule_engine', 
    'stalled_progress_nudge', 
    'v1.2.0', 
    0.88
);
```

### DIFF
--- a/app/api/recommendations/[learner_id]/route.ts
+++ b/app/api/recommendations/[learner_id]/route.ts
@@ -12,10 +12,7 @@ export async function GET(
     const { learner_id } = await params;
 
     if (!learner_id || learner_id === 'undefined') {
-      return NextResponse.json(
-        { error: 'Missing or invalid learner_id' }, 
-        { status: 400 }
-      );
+      return NextResponse.json({ error: 'Missing or invalid learner_id' }, { status: 400 });
     }
 
     // Insert a way to find if a learner_id is in the database
@@ -41,7 +38,7 @@ export async function GET(
       const recDate = new Date(recommendation.createdAt).getTime();
       const now = new Date().getTime();
       const twentyFourHours = 24 * 60 * 60 * 1000;
-      
+
       if (now - recDate > twentyFourHours) {
         isOld = true;
       }
@@ -68,14 +65,10 @@ export async function GET(
       recommended_title: ruleBasedRec.recommendedTitle,
       rationale: ruleBasedRec.rationale,
       source: ruleBasedRec.source,
-      rule_matched: ruleBasedRec.ruleMatched
+      rule_matched: ruleBasedRec.ruleMatched,
     });
-
   } catch (error) {
     console.error('Database Fetch Error:', error);
-    return NextResponse.json(
-      { error: 'Internal Server Error' }, 
-      { status: 500 }
-    );
+    return NextResponse.json({ error: 'Internal Server Error' }, { status: 500 });
   }
 }
--- a/app/api/recommendations/[learner_id]/route.ts
+++ b/app/api/recommendations/[learner_id]/route.ts
@@ -19,7 +19,7 @@ export async function GET(
       );
     }
 
-    /*const [learnerExists] = await db
+    const [learnerExists] = await db
       .select()
       .from(schema.learnerProgress)
       .where(eq(schema.learnerProgress.learnworldsUserId, learner_id))
@@ -30,7 +30,7 @@ export async function GET(
         { error: 'Learner not found' }, 
         { status: 404 }
       );
-    }*/
+    }
 
     const [recommendation] = await db
       .select({
--- a/app/api/recommendations/[learner_id]/route.ts
+++ b/app/api/recommendations/[learner_id]/route.ts
@@ -1,13 +1,12 @@
 import { NextResponse } from 'next/server';
 import { db } from '@/lib/db';
-import { learnerRecommendations, users } from '@/lib/db/schema';
+import { learnerRecommendations } from '@/lib/db/schema';
 import { eq, desc } from 'drizzle-orm';
 import { generateRuleBasedRecommendation } from '@/lib/services/recommendation-model';
-import * as schema from '@/lib/db/schema';
 
 export async function GET(
   request: Request,
-  { params }: { params: { learner_id: string } }
+  { params }: { params: Promise<{ learner_id: string }> }
 ) {
   try {
     const { learner_id } = await params;
@@ -19,18 +18,7 @@ export async function GET(
       );
     }
 
-    const [learnerExists] = await db
-      .select()
-      .from(schema.learnerProgress)
-      .where(eq(schema.learnerProgress.learnworldsUserId, learner_id))
-      .limit(1);
-
-    if (!learnerExists) {
-      return NextResponse.json(
-        { error: 'Learner not found' }, 
-        { status: 404 }
-      );
-    }
+    // Insert a way to find if a learner_id is in the database
 
     const [recommendation] = await db
       .select({
--- a/app/api/recommendations/[learner_id]/route.ts
+++ b/app/api/recommendations/[learner_id]/route.ts
@@ -1,18 +1,17 @@
 import { NextResponse } from 'next/server';
-import { createClient } from '@/lib/supabase/server';
 import { db } from '@/lib/db';
 import { learnerRecommendations } from '@/lib/db/schema';
 import { eq, desc } from 'drizzle-orm';
 
-export async function GET() {
+export async function GET(
+  request: Request,
+  { params }: { params: { learner_id: string } }
+) {
   try {
-    const supabase = await createClient();
-    const {
-      data: { user },
-    } = await supabase.auth.getUser();
+    const { learner_id } = await params;
 
-    if (!user) {
-      return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+    if (!learner_id) {
+      return NextResponse.json({ error: 'Missing learner_id' }, { status: 400 });
     }
 
     const [recommendation] = await db
@@ -24,7 +23,7 @@ export async function GET() {
         rationale: learnerRecommendations.rationale,
       })
       .from(learnerRecommendations)
-      .where(eq(learnerRecommendations.learnworldsUserId, user.id))
+      .where(eq(learnerRecommendations.learnworldsUserId, learner_id))
       .orderBy(desc(learnerRecommendations.createdAt))
       .limit(1);
 
@@ -32,7 +31,7 @@ export async function GET() {
       // Fallback
       return NextResponse.json({
         id: '1d8ca39d-4318-4365-8fe0-293e790495eb',
-        learner_id: user.id,
+        learner_id: 'DEFAULT',
         recommended_item_id: 'course-default',
         recommended_title: 'Default Title',
         rationale: 'Default Rationale',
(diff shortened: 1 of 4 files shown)


zsin520 (changes requested):
Hey Connor,

Thank you for completing this for the recommendation model. Here is my review for the app/api/recommendations/route/ts. 

## FR-9 specifications
1. The endpoint does not fulfill the requirements of `FR-9`. The system should expose a REST endpoint that accepts learner_id and returns a structured JSON response containing the recommendation and rationale. The `export async function GET()` doesn't accept any params. Instead of getting the learner ID from the URL, it extracts it from the Supabase session `(user.id)`. Need to add `learner_id` at `/api/recommendations/[learner_id]`.
2. Need to return 404 for non-existent learners. One of the specifications of `FR-9` is to return 404 for non-existent learners and 400 for missing learner id. 

## Architecture
The endpoint reads from the database, but completely skips `generateRuleBasedRecommendation`. If a user doesn't have a recommendation it falls to a default. I wonder if the endpoint should be invoking the `generateRuleBasedRecommendation` to try and generate a recommendation when one doesn't exist, or if it should invoke a controller that decides whether `generateRuleBasedRecommendation` should be called or the ML recommendation model. 
- If `generateRuleBasedRecommendation` is called from this endpoint then it must also accept `rule_matched` and `source`, which are returned in the recommendation object. Both of these parameters are required for `IC-25`. 
- If not, we may need a controller function to decide the flow recommendations when this endpoint is trying to display a new recommendation for a learner.

app/api/recommendations/[learner_id]/route.ts
```diff
@@ -0,0 +1,93 @@
+import { NextResponse } from 'next/server';
+import { db } from '@/lib/db';
+import { learnerRecommendations, users } from '@/lib/db/schema';
```
Comment (8bit-Apollo) [inline app/api/recommendations/[learner_id]/route.ts]: "users is called but never used" - this is one of the reasons lint check is failing. Remove 'users' here to fix.

app/api/recommendations/[learner_id]/route.ts
```diff
@@ -0,0 +1,93 @@
...
+import { eq, desc } from 'drizzle-orm';
+import { generateRuleBasedRecommendation } from '@/lib/services/recommendation-model';
+import * as schema from '@/lib/db/schema';
+
+export async function GET(
+  request: Request,
+  { params }: { params: { learner_id: string } }
```
Comment (8bit-Apollo) [inline app/api/recommendations/[learner_id]/route.ts]:
```suggestion  
  { params }: { params: Promise<{ learner_id: string }> }
```
I believe since we're using Next.js v15 we need to call params as a promise here.

app/api/recommendations/[learner_id]/route.ts
```diff
@@ -0,0 +1,93 @@
...
+    const { learner_id } = await params;
+
+    if (!learner_id || learner_id === 'undefined') {
+      return NextResponse.json(
+        { error: 'Missing or invalid learner_id' }, 
+        { status: 400 }
+      );
+    }
+
+    const [learnerExists] = await db
```
Comment (8bit-Apollo) [inline app/api/recommendations/[learner_id]/route.ts]: This checks if the learner has an entry in the progress table not if they exist. Because of this, a user with no progress will return 404 instead of a default recommendation as required by FR-8.

Instead of checking the learnerProgress table, I'd recommend checking the learnworlds learner table to see if the user exists... But it doesn't look like that's implemented (at least on our end). 

Perhaps we remove this check and add a comment to let the Learnworld's team know to add it when they take over?

@sserurich Your thoughts on this would be appreciated.

app/api/recommendations/feedback/route.ts
```diff
@@ -1,10 +1,48 @@
...
+    if (!user) {
+      return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+    }
 
-  return NextResponse.json({ message: 'Feedback received' }, { status: 200 });
+    const body = await request.json();
+
+    const { recommendationId, learnworldsUserId, recommendedItemId, feedbackType, rating, comment } = body;
+
+    if (!recommendationId || !recommendedItemId) {
```
8bit-Apollo: We should validate all required fields before inserting feedback. Right now the route only checks recommendationId and recommendedItemId, but it also inserts learnworldsUserId, feedbackType, rating, and comment.

8bit-Apollo: might also be prudent to check if feedbackType and Rating are within accepted parameters.

app/api/recommendations/feedback/route.ts
```diff
@@ -1,10 +1,48 @@
...
+      return NextResponse.json(
+        { error: 'Missing required fields: recommendationId or recommendedItemId' },
+        { status: 400 }
+      );
+    }
+
+    const [newFeedback] = await db
```
8bit-Apollo, line 70: Before inserting, can we verify that the `recommendationId` exists and belongs to the same `learnworldsUserId` / `recommendedItemId`? Otherwise feedback could be attached to an unrelated or invalid recommendation.

components/rec-widget.tsx
```diff
@@ -35,14 +28,32 @@ export function RecommendationWidget() {
...
+          
+          if (error) throw error;
+          
+          learnerId = data.user?.user_metadata?.learnworlds_id;
+          console.log("Successfully inserted mock ID:", learnerId);
         }
 
-        await new Promise((resolve) => setTimeout(resolve, 800));
-        setData(MOCK_DATA);
+        const response = await fetch('/api/recommendations/' + user.id);
```
8bit-Apollo: It looks like this component gets `learnerId` from `user_metadata.learnworlds_id`, but the recommendation request still uses `user.id`. If `/api/recommendations/[learner_id]` expects the LearnWorlds learner ID, can we use `learnerId` here instead of the Supabase auth user ID?

components/rec-widget.tsx
```diff
@@ -51,36 +62,31 @@ export function RecommendationWidget() {
...
     try {
       const response = await fetch('/api/recommendations/feedback', {
         method: 'POST',
         headers: { 'Content-Type': 'application/json' },
         body: JSON.stringify({
-          learner_id: userId || data?.learner_id || 'ERROR FETCHING ID',
-          recommendation_id: data?.recommended_item_id,
-          rating_value: rating,
-          timestamp: new Date().toISOString(),
+          recommendationId: data.id,
```
8bit-Apollo, line 79: I think we need a check to make sure recommendationId is not null, here.

8bit-Apollo (changes requested):
Hey @Connor-Palmira, thanks for you work on this PR. There were a few issues I saw that need to be addressed before this is ready. Could you take a look at the comments and update the implementation accordingly? Once those are handled, this should be good to merge.

app/api/recommendations/[learner_id]/route.ts
Comment (Connor-Palmira) [reply app/api/recommendations/[learner_id]/route.ts]: Resolved: Removed `users`

app/api/recommendations/[learner_id]/route.ts
Comment (Connor-Palmira) [reply app/api/recommendations/[learner_id]/route.ts]: Resolved: Remove the check and added a comment that asks for it to be added in future updates

app/api/recommendations/feedback/route.ts
Connor-Palmira: Resolved: Added validation for all required fields (does not include comment as that is optional). Also added the check to see if feedbackType and rating are accepted values.

components/rec-widget.tsx
Connor-Palmira: Resolved: Replaced `user.id` with `learnerId`

components/rec-widget.tsx
Connor-Palmira: Resolved: Added a check to make sure `recommendationId` is not null

app/api/recommendations/[learner_id]/route.ts
Comment (Connor-Palmira) [reply app/api/recommendations/[learner_id]/route.ts]: Resolved: Added Promise when calling for params

app/api/recommendations/feedback/route.ts
Connor-Palmira: Resolved: Added a check that the learner must have that recommendation for them to give feedback

sserurich:
thanks @Connor-Palmira for addressing the items identified. Most of the inline comments have been addressed, especially the validation, route params, learnerId usage, and recommendation ownership checks. However, the PR still does not appear to implement the requested 404 behavior for non-existent learners. This will make the system crush incase a valid learner id is passed but not available.  you can consider adding something like this
```js 
// 3. The "Not Present" check
if (!learner) {
  return NextResponse.json({ error: 'Learner not found' }, { status: 404 });
}```

tylerbadcock:
Closed based on 16/06/2026 meeting and moving away from LearnWorlds integration.

PR #170, opened 2026-04-20
added reranking recommendation model using a cross encoder

### LINKED ISSUE #141: IC-39 feat: implement ML inference pipeline for recommendation scoring
### Description
Use trained model to score candidate items and select the best recommendation.

### Scope
- Load active model  
- Score candidates  
- Select top item  
- Use features derived from learner state and `learner_item_events`

### Acceptance Criteria
- Model scores candidates correctly  
- Top item is selected deterministically  
- Model version tracked  
- Interaction features are consistent with training

lolbro2k:
### Description
Using BAAI/bge-reranker-base (Cross encoder) from Xenova/bge-reranker-base as the quantized 16bit model, We more accurate rerank the recommendation with less susceptibility to buzzwords. Furthermore, the reranker scores high on relevancy testest much more than cosine similarity in BAAI/bge-m3 (original AI pre-screening), so much more accurate courses are recommended. The input to this function is AI screen courses (only the top AI pre-screening courses) and the current course description. The reranker then reranks the course description and return the most relevant courses.

### Architectural and Model Notes
This is just a smaller portion of the entire model. First the bge-M3 is used as AI prescreening, This prescreen was done in a different PR to vectorize the description then cosine similarity. Cosine similarity is too broad and is easily confused with repeated "buzz words". I implemented this as a secondary recommendation that reranks only the top scoring courses from the previous pre-screening PR. This is because Cross-encoder are much more expensive, thus I used a quantized version too, to rerank the courses. The additional PR must be implemented to incorporate clicks, time-spent, and other metrics by the users. A single PR for all of those metrics is too big, so this PR focuses on solely the reranking of only the top courses, as part of the entire Model recommendation architecture.

### Scope
- [x] Load active model  
- [x] Score candidates  
- [x] Select top item  
- [x] Use features derived from learner state and `learner_item_events`
### Not within scope
- [] Entire pipeline to feed then recommend courses, this logic is for another issue to create the pipeline of the entire recommendation system

### Acceptance Criteria
- [x] Model scores candidates correctly  
- [x] Top item is selected deterministically  
- [x] Model version tracked  
- [x] Interaction features are consistent with training  

### Testing
- [] Testing insured with fake course data still need to be completed, but that is rewritten for another issue.

### Note
works for chinese courses now too

### DIFF
--- a/lib/services/ML-recommendation-Model.ts
+++ b/lib/services/ML-recommendation-Model.ts
@@ -45,7 +45,7 @@ export async function CrossEncoderRecommender(
       truncation: true,
     });
     const output = await model(inputs);
-    const score = output.logits.data[0];
+    const score: number = Number(output.logits.data[0]);
     scoredCourses.push({
       course: candidateCourseDescription,
       score: score,
--- a/lib/services/ML-recommendation-Model.ts
+++ b/lib/services/ML-recommendation-Model.ts
@@ -1,0 +1,57 @@
+import {
+  AutoModelForSequenceClassification,
+  AutoTokenizer,
+  PreTrainedModel,
+  PreTrainedTokenizer,
+} from '@huggingface/transformers';
+
+const RULES_CONFIG = {
+  TOP_COURSES: 3,
+};
+
+// load model in when server starts. Using BAAI/bge-reranker-base with Xenova version
+// BAAI/bge-reranker-base with 16b quantized for lightweight
+class CrossEncoderPipeline {
+  static modelId = 'Xenova/bge-reranker-base';
+  static tokenizer: PreTrainedTokenizer | null = null;
+  static model: PreTrainedModel | null = null;
+
+  static async getInstance() {
+    if (this.tokenizer === null || this.model === null) {
+      this.tokenizer = await AutoTokenizer.from_pretrained(this.modelId);
+      this.model = await AutoModelForSequenceClassification.from_pretrained(this.modelId);
+    }
+    return { tokenizer: this.tokenizer, model: this.model };
+  }
+}
+
+/**
+ * Re-ranks a list of candidate courses based on how similar they are to a completed course.
+ * @param candidateCourses using the BAAI/bge-M3 vectorized version, Submit only the top scoring courses
+ * @param completedCourseDescription The description of the course the user just completed or is looking at
+ * returns a reranked recommendation lists using BAAI/bge-reranker-base returning top 3 recommendation
+ */
+export async function CrossEncoderRecommender(
+  candidateCourses: string[],
+  completedCourseDescription: string
+) {
+  const { tokenizer, model } = await CrossEncoderPipeline.getInstance();
+
+  const scoredCourses = [];
+  for (const candidateCourseDescription of candidateCourses) {
+    const inputs = await tokenizer(completedCourseDescription, {
+      text_pair: candidateCourseDescription,
+      padding: true,
+      truncation: true,
+    });
+    const output = await model(inputs);
+    const score = output.logits.data[0];
+    scoredCourses.push({
+      course: candidateCourseDescription,
+      score: score,
+    });
+  }
+  scoredCourses.sort((a, b) => b.score - a.score);
+
+  return scoredCourses.slice(0, RULES_CONFIG.TOP_COURSES);
+}
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,5 +1,5 @@
 {
-  "name": "judgeportal",
+  "name": "judge-portal",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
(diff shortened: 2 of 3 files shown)


lib/services/ML-recommendation-Model.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  const scoredCourses = [];
+  for (const candidateCourseDescription of candidateCourses) {
+    const inputs = await tokenizer(completedCourseDescription, {
+      text_pair: candidateCourseDescription,
+      padding: true,
+      truncation: true,
+    });
+    const output = await model(inputs);
+    const score = output.logits.data[0];
```
8bit-Apollo: Might be safer to explicitly convert this. As is, there's no guarantee  output.logits.data[0] will point to a plain numeric value.

lib/services/ML-recommendation-Model.ts
```diff
@@ -0,0 +1,57 @@
...
+      this.tokenizer = await AutoTokenizer.from_pretrained(this.modelId);
+      this.model = await AutoModelForSequenceClassification.from_pretrained(this.modelId);
+    }
+    return { tokenizer: this.tokenizer, model: this.model };
+  }
+}
+
+/**
+ * Re-ranks a list of candidate courses based on how similar they are to a completed course.
+ * @param candidateCourses using the BAAI/bge-M3 vectorized version, Submit only the top scoring courses
```
8bit-Apollo, line 30: BGE-M3? or are we using the Reranker here?

8bit-Apollo (commented):
Good Implementation. Other than the two issues noted, everything looks good. May be prudent to switch to batch scoring long term, but that's an issue for future PRs. For now, I believe this is ready to merge (pending the changes I suggested).

zsin520 (commented):
Hey Mathew, thank you for working on this important feature of the project. I have several comments for the architecture of the model.

## Fundamental behavior of the model
1. The model takes the course that was just completed and runs the algorithm to compare it against several other candidate courses, returning the course with the best score.
2. A big piece of the model is to consume the user behavior as well as the course descriptions. We can assume there is enough information for the user since it was directed to the ML recommender, so it should consider behavior like views, completions, and clicks. 
3. This may be out of the scope of this PR and this cross encoder may just focus on the course descriptions and serve as part of the recommender model. If this is the case it should be clear that this is only a piece of the model. 

## Model input and output 
1. Currently the model accepts `completedCourseDescription` as a string and returns `candidateCourseDescription` with the best score. The PR says "The input to this function is AI screen courses (only the top AI pre-screening courses) and the current course description". It sounds like there is another process that pre-filters all the possible courses? Is this AI course screening implemented somewhere else ? 
2. When the API is called for the model, I believe it should be returning a course code back to LearnWorlds. Otherwise there is no way for LearnWorlds to link the user to the course and there is no safe way to identify the correct course code based simply on the course description (this would also add overhead). 
3. I think we should consider using the same architecture as PR #156, the rule-based recommendation engine. It accepts the learners ID, which can easily be passed from LearnWorlds. The course information can be extracted from there. In addition, the rule-based model builds a `RecommendationResult` object, which is a good way to standardize output from our models. This also includes the rationale for the model's decision, which is a requirement of the FR.

## Recommendation questions
- Right now it seems like the model will recommend any course with a good score. I wonder if we should consider two types of recommendations. The first recommendation should be the next course in the series. 
- The model's recommendation may be invoked if the user is interested in "more courses like this", or if they complete a series.
- Let me know your thoughts on this. 

## Summary
1. Decide whether this is part of the final ML model, or if it is the final ML model. (Considering user behavior in the recommendation may be implemented later on). 
2. Standardize the input and output of the model to match the rule-based model and better integrate with Supabase and LearnWorlds.

lib/services/ML-recommendation-Model.ts
lolbro2k: The BAAI/bge-M3, is the prescreen AI that Jacob implemented. This is built on top of that pre-screening, because cross encoders are much more computationally expensive, so its input is the prescreened bge-M3. This allows the cross encoder to only rerank a smaller input batch of courses

8bit-Apollo: Ah ok, thanks for the clarification. That should be fine, just keep in mind the BGE-M3 implementation currently only supports scoring challenges (criteria vs submissions). A new function will have to be created to handle Courses.

sserurich:
@lolbro2k , thanks for the work on this.

You’re close, but a couple review items still need follow-through before I can approve:

- Inline fix still missing: please make output.logits.data[0] an explicit JS number (and ensure sorting is numeric-safe), per the earlier comment.
- Architecture/IO feedback not addressed: this currently takes/returns course descriptions; for LearnWorlds/Supabase integration we likely need standardized input/output (e.g., align with PR #156 / RecommendationResult, return a course identifier + rationale). If that’s intentionally out-of-scope, please document it clearly in the PR and/or open a follow-up issue.
- Dependency note: adding @huggingface/transformers pulls in heavy runtime deps (onnx/sharp, etc.); please confirm this is intended for the server runtime and won’t break deploys.
-  Lastly the git actions (CI) is failing , update with the latest main branch and ensure all these pass

tylerbadcock:
Closed based on 16/06/2026 meeting. Noted the potential for security issues with dependencies.

PR #172, opened 2026-04-21
fix: resolve participant redirect loop and stabilize mentor testing

AngelVargas1009:
I resolved an issue where participant users were stuck in a redirect loop between / and /participant, which prevented access to the mentor search page.

The root cause was inconsistent role handling between middleware and the participant layout. Middleware normalized roles, while the layout compared raw role values, causing mismatched redirects.

Changes made:

Normalized user role in app/participant/layout.tsx using trim().toLowerCase()
Ensured consistent role checks between middleware and layout
Verified participant users can now successfully access /participant and /participant/mentors

This fix allows proper testing of the mentor search/filter feature and ensures stable participant routing.

This PR is a stabilization/testing branch used to validate mentor filtering and participant routing behavior after earlier mentor search implementation work.


Connor-Palmira (commented):
This is a solid PR that resolves the redirect loop and stabilizes testing for mentors. This PR was done really well but there were some things that I noticed that brought up some questions. This PR is for resolving the redirect loop but in `tests/mentor-search.spec.ts` you still have `test.skip` set to true due to `Skipping mentor search Playwright tests due to auth middleware redirect in the test environment`, which sounds like it should be fixed by this pull request, so it should no longer be set to true.

Additionally, make sure to run `npx prettier --write <file-path>` for the files `app/participant/layout.tsx` and `lib/supabase/middleware.ts`.

Overall, this is a really solid PR.

AngelVargas1009:
Latest main has been merged and lint/format is now passing.

The remaining Playwright failure is coming from the LearnWorlds admin API tests introduced from main, not from the mentor routing/testing changes in this PR. The failing tests expect authenticated non-admin users to receive 401 on `/api/admin/learnworlds/ingest` and `/api/admin/learnworlds/transform`, but the response is currently 200.

tylerbadcock:
Couldn't seem to replicate this one on main. Maybe it was already fixed in another PR? Also worth noting that the participant role is already lowercase, so it seems this error may never be encountered?

tylerbadcock:
Closed based on 23/06/2026 meeting.

PR #176, opened 2026-04-23
MD:14 Mentor Seed Data

### LINKED ISSUE #175: MD-14: Mentor Seed Data
Create a seed dataset to simulate a populated database of mentor profiles.

Correspond with the `mentor_profiles` table:
- `id`
- `learnworlds_user_id`
- `full_name`
- `title`
- `organization`
- `bio`
- `linkedin_url`
- `calendly_url`
- `photo_url`
- `tags`
- `is_visible`
- `created_at`
- `updated_at`

Connor-Palmira:
This PR creates seed data that populates the `mentorProfiles` table with the mock data that was displayed during the demo. This data is used to simulate a populated database without any real data.

Changes
- Updates `scripts/consolidated-seed.ts` with sample mentor data
- Start message `Seeding mentor profiles...` and success message `- Mentors: ${sampleMentors.length}`

Closes #175


lolbro2k (changes requested):
This PR focus only solely seed data, I think, should have labels or notes next to the actual data to notify this. The seed data, when used in tandem with actual data, may get confusing in future PRs and implementation. The usage of adding a keyword next to the organization or name would be useful to indicate this data is not real-world data. In a sql table it would all look the same between seed data and real-worlds

### Changes Requested
1. Include some descriptive factor to note this data is seed data such that in use you can differentiate between which was was seed data and real-world. (e.g. add "-seed" to organization, add "seed-..." to title, include "seed" in bio or ...)
2. run npm run lint & npm run format again please.

nsoliven (changes requested):
I pulled this branch locally and checked the seed script. The mentor seed data itself looks fine, and the `SEED` organization prefix seems to address the earlier feedback about marking demo data.

One actual blocker: this new insert uses `mentorProfiles`, but `mentorProfiles` is not imported at the top of the file. I ran a typecheck-style check and it reports:

`lib/db/seed.ts(372,21): Cannot find name 'mentorProfiles'.`

Since this PR is specifically adding mentor seed data, `npx tsx lib/db/seed.ts` cannot reliably run the new mentor seed block until this is fixed. Can we add `mentorProfiles` to the schema import?

sserurich (commented):
**Problem Summary**  
The seed script was failing with `database doesn't exist` errors.
It needs to be fixed before merge

sserurich (approved):
Thanks @Connor-Palmira , @nsoliven , I confirm this now works after updating /resolving the earlier issue with the new approach we discussed.

Issue #164, opened 2026-04-17
IC-50 feat: Create global service health monitoring endpoint


Thomas12556:
Establish a standardized /health endpoint to monitor overall service availability for the Judge Portal, including database connectivity and the status of the LearnWorlds ingestion pipeline.

Acceptance Criteria
---

- [x] A GET request to /api/health returns the aggregated JSON payload with a valid ISO timestamp. 
- [x] If the database query fails, returns status: "unavailable" and database: "disconnected" with a 503 HTTP code.

Dependencies
---

-  lib/learnworlds/health.ts (from IC-21 pull request)

PR #123, opened 2026-04-10
IC-21 feat: add ingestion health monitoring and staleness check

zsin520:
Closes IC-21, related to FR-15 Service Health Monitoring Endpoint
## Summary
This PR introduces a health monitoring utility for the LearnWorlds ingestion pipeline. By querying the `learnworlds_sync_runs` table via Drizzle, it determines if the data feeding our recommendation engine is up-to-date, failing, or stale. It also exposes an admin-secured API endpoint to integrate with our broader FR-15 service health monitoring.

## Changes
- `lib/learnworlds/health.ts`: Core logic for checking sync run recency and status. Defines `IngestionHealthReport` which returns a status, message, the last run time, and error details. Checks the most recent ingestion in the `learnworlds_sync_runs` table and returns the status of the pipeline. 
- app/api/admin/learnworlds/health/route.ts: Admin endpoint to expose the health status. Checks the user is an admin and waits for the health report from `getIngestionHealthStatus`. Returns health report and http status. Utilizes `sendApiError` for errors.

## Notes
- Dependencies: PR #103  for `sendApiError` utility. PR #97 for `learnworldsSyncRuns` table.

PR #177, opened 2026-04-28
feat(IC-50): Implement Global Health Check Endpoint

Thomas12556:
Description 
---
Closes: (IC-50) #164 
Fulfills: FR-15 (Service Health Monitoring Endpoint)

This PR adds a centralized `/api/health` endpoint to monitor the Judge Portal's status. It aggregates database connectivity and the LearnWorlds ingestion subsystem into a single diagnostic response.

Overview
---
1. Database: Executes `SELECT 1` to verify the connection. Returns `503` if disconnected.

2. Ingestion: Calls `getIngestionHealthStatus()`. If the sync is failing or stale, the global status becomes degraded but stays `200`.

3. Fault Tolerance: Errors in the ingestion subsystem are caught and reported as degraded rather than crashing the health check.

Dependencies
---
PR #123 implementation of `getIngestionHealthStatus()`.

Sample Response
---
```json
{
  "status": "healthy",
  "database": "connected",
  "timestamp": "2026-04-27T...",
  "subsystems": {
    "ingestion": {
      "status": "healthy",
      "lastRunAt": "..."
    }
  }
}
```

### DIFF
--- a/app/api/health/route.ts
+++ b/app/api/health/route.ts
@@ -5,7 +5,7 @@ import { getIngestionHealthStatus } from '@/lib/learnworlds/health';
 
 export async function GET() {
   const timestamp = new Date().toISOString();
-  
+
   let dbStatus = 'connected';
   let globalStatus = 'healthy';
   let httpStatus = 200;
@@ -28,7 +28,7 @@ export async function GET() {
   if (dbStatus === 'connected') {
     try {
       const ingestionReport = await getIngestionHealthStatus();
-      
+
       ingestionStatus = {
         status: ingestionReport.status,
         lastRunAt: ingestionReport.lastRunAt,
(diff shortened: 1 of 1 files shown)


Connor-Palmira (commented):
This PR is a great implementation of a health check endpoint. You handle the state of `globalStatus` and `dbStatus` well. There are a few suggestions I have that you may want to consider.

When you call the database with `SELECT 1` you may want to add some sort of timeout so that the health check does not hang indefinitely.

When an error occurs, you set `ingestionStatus.status` to `unknown`, but you may want that to be set to `unavailable` instead because the health check failed.

Overall, this is a really good implementation and I think once you address these concerns of mine this PR will be ready for merging.

Thomas12556:
Thanks for the feedback, Connor. I've pushed an update addressing both points:

1. Wrapped the database `SELECT 1` execution in a Promise.race with a 5 second timeout to prevent indefinite hanging.

2. Updated the catch block for the ingestion check to flag ingestionStatus.status as unavailable rather than unknown on failure.

sserurich (commented):
@Thomas12556  , this fails when I run locally. 

Image

I have checked, the dependant PR ([https://github.com/Freddie-Pike/judge-portal/pull/123](https://github.com/Freddie-Pike/judge-portal/pull/123) isn't yet merged so can't merge in this, it will create issues if I merge this in.

tylerbadcock:
This PR uses LearnWorlds integration and relies on dependencies from PR #123 which was closed. Based on this and the meeting on 23/06/2026, I have closed this PR for now.

Issue #169, opened 2026-04-20
Refactor API Routes to use Standardized sendApiError Function


zsin520:
## Background 
The standardized API error handling function `sendApiError` (satisfying FR-14) has been merged into main and is now available at `/judge-portal/lib/utils/api-errors.ts`.

We need to update all existing API endpoints to replace custom NextResponse.json() error returns with this new centralized utility. This will guarantee that all endpoints return the exact structured JSON format for errors.

## Implementation Strategy
To minimize merge conflicts and ensure thorough review, every file change will be submitted as its own separate Pull Request. 
1. Import the utility:
`import { sendApiError } from '@/lib/utils/api-errors';`
2. Example of changing error return:
`return NextResponse.json({ error: 'Internal server error' }, { status: 500 });
// OR
return sendApiError(500, 'INTERNAL_SERVER_ERROR', 'Internal server error');`

## Target files
There are several groups of route files that need to be updated: admin routes, super admin routes, judge routes, Participant & Challenge Routes, and General / Auth Routes.

PR #178, opened 2026-04-30
Refactor: Implement sendApiError in Admin Criteria Routes

Thomas12556:
Implements part of #169

Description
---

This PR refactors the error handling in the Admin Criteria API routes to comply with FR-14: Basic Error Handling.

Changes
---
- Migrated from NextResponse.json() calls to the sendApiError utility.
- Replaced `NextResponse.json({ error: ... })` with `sendApiError(status, code, message)`.
  - The standardized error codes used are the following:
 
| Error Name | Status Code |
| :--- | :--- |
| `MULTIPLE_CHOICES` | 300 |
| `BAD_REQUEST` | 400 |
| `UNAUTHORIZED` | 401 |
| `FORBIDDEN` | 403 |
| `NOT_FOUND` | 404 |
| `CONFLICT` | 409 |
| `UNPROCESSABLE_ENTITY` | 422 |
| `INTERNAL_SERVER_ERROR` | 500 |

Updated routes:

1. `app/api/admin/criteria/route.ts`

2. `app/api/admin/criteria/reorder/route.ts`

3. `app/api/admin/criteria/[criterionId]/route.ts`

### DIFF
--- a/app/api/admin/criteria/[criterionId]/route.ts
+++ b/app/api/admin/criteria/[criterionId]/route.ts
@@ -4,6 +4,7 @@ import { db } from '@/lib/db';
 import { criteria } from '@/lib/db/schema';
 import { eq } from 'drizzle-orm';
 import { getAdminOrgId, requireEventInOrg } from '@/lib/auth/org';
+import { sendApiError } from '@/lib/utils/api-errors';
 
 export async function PUT(
   request: NextRequest,
@@ -16,7 +17,7 @@ export async function PUT(
     const user = await getUserFromSession();
 
     if (!user || user.role !== 'admin') {
-      return NextResponse.json({ error: 'Unauthorized' }, { status: 401 });
+      return sendApiError(401, 'UNAUTHORIZED', 'Unauthorized');
     }
 
     const orgId = await getAdminOrgId(user.id);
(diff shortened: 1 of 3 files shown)


Connor-Palmira (approved):
This is a nice and clean refactor that moves away from `NextResponse` and adopting `sendApiError`. From what I can tell, everything looks good. The consistency of using standardized error codes like `BAD_REQUEST` and `UNAUTHORIZED` is really nice and implemented well. Overall, it looks easy to read and you have done a good job.

Great work with the refactor, I think this is good for merging.

app/api/admin/criteria/route.ts
```diff
@@ -59,50 +60,44 @@ export async function POST(request: NextRequest) {
...
     const { eventId, name, description, minScore, maxScore, weight, category } =
       await request.json();
 
     if (!eventId) {
-      return NextResponse.json({ error: 'Event ID is required' }, { status: 400 });
+      return sendApiError(400, 'BAD_REQUEST', 'Event ID is required');
     }
 
     const orgId = await getAdminOrgId(user.id);
     await requireEventInOrg(eventId, orgId);
```
8bit-Apollo: If `requireEventInOrg(...)` throws when the event is missing or outside the admin’s organization, this will fall into the generic catch block and return a 500. Since that would be a normal access/not-found case, should we catch that and return a standardized 403 or 404 with `sendApiError(...)` instead?

app/api/admin/criteria/[criterionId]/route.ts
```diff
@@ -165,7 +157,7 @@ export async function DELETE(
       .limit(1);
 
     if (!existingCriterion) {
-      return NextResponse.json({ error: 'Criterion not found' }, { status: 404 });
+      return sendApiError(404, 'NOT_FOUND', 'Criterion not found');
     }
 
     await requireEventInOrg(existingCriterion.eventId, orgId);
```
Comment (8bit-Apollo) [inline app/api/admin/criteria/[criterionId]/route.ts]: Same concern as below: if `requireEventInOrg(...)` throws for an org mismatch or missing event, this route will return a generic 500. Should this be mapped to a standardized 403 or 404 instead so org-scope failures do not look like internal server errors?

app/api/admin/criteria/reorder/route.ts
```diff
@@ -4,36 +4,36 @@ import { db } from '@/lib/db';
...
 
     const orgId = await getAdminOrgId(user.id);
     const { eventId, criteriaOrders } = await request.json();
 
     if (!eventId) {
-      return NextResponse.json({ error: 'Event ID is required' }, { status: 400 });
+      return sendApiError(400, 'BAD_REQUEST', 'Event ID is required');
     }
 
     await requireEventInOrg(eventId, orgId);
```
8bit-Apollo: Here too

8bit-Apollo (dismissed):
Overall this looks good for FR-14. The criteria routes now consistently use `sendApiError(...)` for unauthorized, validation, not found, duplicate constraint, and fallback server errors, which is a solid improvement over mixed response shapes.

The main thing I would like confirmed is how `requireEventInOrg(...)` behaves. If it throws for an org mismatch or missing event, these routes currently fall into the generic catch block and return a 500. I think those should probably be mapped to a standardized 403 or 404 so normal org-scope/access failures do not look like internal server errors.

Other than that, this looks aligned with the FR-14 direction.

Thomas12556:
Thank you for the reviews @Connor-Palmira and @8bit-Apollo. Regarding the mapping of `requireEventInOrg(...)` to specific 403 or 404 codes, I believe this is slightly out of scope for this PR. I've looked into the implementation and it seems to throw a generic error without an error code to differentiate between 403 or 404, therefore I would need to reimplement this function. 

```ts
// Implementation in /lib/auth/org.ts:
export async function requireEventInOrg(eventId: string, orgId: string): Promise<void> {
  const belongs = await verifyEventInOrg(eventId, orgId);
  if (!belongs) {
    throw new Error('Event does not belong to your organization');
  }
}
```
However, this deviates away from the purpose of this PR which was to focus on refactoring the `NextResponse.json()` error returns. I believe we should open up a new issue regarding this.

tylerbadcock (approved):
Approved based on 30/06/2026 meeting.

tylerbadcock (approved):

PR #185, opened 2026-05-01
IC-40 feat: implement recommendation orchestrator

zsin520:
## Description
IC-40, IC-26 feat: implement hybrid recommendation orchestrator (rules + ML). This PR implements the 'getRecommendation' orchestrator which performs freshness check and decides which recommendation model should be invoked. 

## Summary
1. Implements freshness check for `IC-26`. If a valid recommendation exists within the `FRESHNESS_HOURS_THRESHOLD`, it is reused to save compute and API calls.
2. Checks `learnerItemEvents` for how many events are recorded to decide whether the ML model recommendation may be called. Uses `ML_READY_EVENT_THRESHOLD` currently set to 5.
3. Learners with little to no data use the rule-based model `generateRuleBasedRecommendation`.
4. Adds `persistRecommendation` to ensure recommendations are properly logged.

## Notes
A new file `lib/services/ml-based-recommendation-model.ts` was also created, which will hold code for ML model recommendation, but is currently acting as a placeholder since this has not been implemented yet.


zsin520:
This code has been added to PR#156, so this PR will be closed.

PR #206, opened 2026-05-25
Add TypeScript GitHub Action

### LINKED ISSUE #204: Add TypeScript Github Action
This task is to add a Github Action for TypeScript linting. A script will need to be added to do a type check.

tylerbadcock:
Closes #204


Freddie-Pike (changes requested):
PR looks great! There's an Github action error as we'll need to address which is a Prettier error.

tylerbadcock:
Prettier formatting should be fixed. I had to update playwright to the latest version as the test was hanging otherwise.

.github/workflows/typecheck.yml
```diff
@@ -0,0 +1,28 @@
+name: TypeScript Check
+
+on:
+  push:
+    branches: ['**']
```
Freddie-Pike: You can remove the on push, and just have this be on pull requests.

Freddie-Pike (approved):
Code reviewed, PR LGTM. Thanks for the action fixes!

